PR #134, opened 2019-04-18
reduces test input for cosim_cuda_graph_degree_test and prints output

max-ruttenberg:
* reduces input for graph degree test


max-ruttenberg:
This PR seems to be dead too.

PR #154, opened 2019-04-23
adds cuda library source files to regression

max-ruttenberg:
The Cuda library was moved into it's own source and header file. This PR updates the regression test suite to account for this.


max-ruttenberg:
@drichmond I'm pretty sure this PR can be killed.

Issue #173, opened 2019-05-03
Read all configuration parameters from hb_mc_get_config


drichmond:
https://github.com/bespoke-silicon-group/bsg_f1/blob/0d415d044fb3ba7ed89a6c7c3b2920ff0dfcbae1/cl_manycore/libraries/bsg_manycore_driver.cpp#L441

At the moment we have several methods for reading the configuration from a device. Let's standardize on using hb_mc_get_config, which will read the configuration at initialization and cache it in a device-descriptor indexed table (so it doesn't need to be re-read on every invocation). 

This will have the added benefit of adding support for multiple manycore FPGAs, which is currently not supported.

drichmond:
Fixed in #213

PR #195, opened 2019-05-10
Refactored get_config to read configurations from table

save-buffer:
Addresses #173

### DIFF
--- a/cl_manycore/libraries/bsg_manycore_driver.cpp
+++ b/cl_manycore/libraries/bsg_manycore_driver.cpp
@@ -1,6 +1,7 @@
 #include <stdlib.h>
 #include <stdio.h>
 #include <string.h>
+#include <array>
 
 #ifndef COSIM
 	#include <bsg_manycore_errno.h> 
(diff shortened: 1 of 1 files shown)


max-ruttenberg:
Is this PR stale?

save-buffer:
I'm not quite sure what the status on whether we need this feature or not is. It definitely needs rebasing.

max-ruttenberg:
What is exactly is the feature?

save-buffer:
To read the ROM from a device when it is initialized, and store the contents in a table indexed by that device's file descriptor. This change is necessary because we just refactored `hb_mc_format_request_packet` to have a call to `get_config`. This call will be very very slow under heavy network load if we don't do this because it will have to go to the rom each time a packet has to be formatted.

save-buffer:
But @drichmond said that we were still deciding whether to include it like this or as part of a much larger refactor.

drichmond:
Yes - Sasha is correct here. I asked him to do this to solve this problem, but our intended refactor that we finalized this morning will make this PR stale.

Issue #202, opened 2019-05-13
Make cpp files in library use c++ headers


save-buffer:
Minor nit, but it should probably be done just to avoid any potential weirdness that might arise from using c headers.

max-ruttenberg:
Do you mean the standard headers like cstdio vs stdio.h?

save-buffer:
Yeah

PR #204, opened 2019-05-14
switched to cpp headers

save-buffer:
Switched to conditionally include the c++ versions of c headers. Passes both cosim and on f1. 
Addresses #202

### DIFF
--- a/cl_manycore/libraries/bsg_manycore_bits.h
+++ b/cl_manycore/libraries/bsg_manycore_bits.h
@@ -5,7 +5,12 @@
 #else
 #include "bsg_manycore_features.h"
 #endif
+
+#ifdef __cplusplus
+#include <cstdint>
+#else
 #include <stdint.h>
+#endif
 
 #ifdef __cplusplus
 extern "C" {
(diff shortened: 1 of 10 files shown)


max-ruttenberg:
@save-buffer 

Have you:
1. Run the regression tests this on F1? Compiles?
2. Run the regression tests this on COSIM? Compiles?

If you have, can you post the output? 

The only reason I can think of that it wouldn't compile might be the need for a `std::` prefix, but I can't remember if that's actually needed. I would think that it isn't but it's good practice to run the regression tests anyway.

save-buffer:
On F1:
image
image

All the tests pass in cosim as well, but after they've all run I'm getting some error about not finding `cadenv.mk`.

max-ruttenberg:
@drichmond have you seen that error before?

drichmond:
Negative

max-ruttenberg:
@save-buffer can you get please email me and @drichmond the full log?

Do you see the same error with just dev checked out (without your changes)?

max-ruttenberg:
@save-buffer it sounds to me like you’re missing a repo that should be in the same directory as bsg_f1.

save-buffer:
Ah, sorry, I didn't see these comments for some reason in my email. I don't have `bsg_cadenv` cloned.

PR #205, opened 2019-05-14
Added Makefile code to check for/include for Bladerunner Makefiles

drichmond:
* Sets BASEJUMP_STL_DIR BSG_IP_CORES_DIR BSG_MANYCORE_DIR environment variables
  from Makefile.environment.

* Overrides previously set environment variables (warns if this occurs)

### DIFF
--- a/cl_manycore/Makefile.environment
+++ b/cl_manycore/Makefile.environment
@@ -6,6 +6,23 @@ ifndef CL_DIR
 $(error "CL_DIR undefined. Set CL_DIR before running this makefile")
 endif
 
+# Default to overriding BSG_IP_CORES_DIR and BSG_MANYCORE_DIR and warn
+ifneq ("$(wildcard ../../Makefile.deps)","")
+ORANGE=\033[0;33m
+NC=\033[0m
+
+ifdef BSG_IP_CORES_DIR
+$(warning $(shell echo -e "\t\n\t$(ORANGE)Overriding BSG_IP_CORES_DIR environment variable with Bladerunner defaults.$(NC)"))
+endif
+
+ifdef BSG_MANYCORE_DIR
+$(warning $(shell echo -e "\t\n\t$(ORANGE)Overriding BSG_MANYCORE_DIR environment variable with Bladerunner defaults.$(NC)"))
+endif
+
+include ../../Makefile.common
+
+endif
+
 ifndef BSG_IP_CORES_DIR
 $(error "BSG_IP_CORES_DIR environment variable undefined.")
 endif


max-ruttenberg:
@drichmond 

What should be done to test this PR?

On F1, with AMI from Bladerunner 0.4.1
I've cloned `bsg_f1` to `/home/centos/bsg_bladerunner/` and checked out this branch.

What I see is this:
```
/home/centos/bsg_bladerunner/bsg_f1/cl_manycore/Makefile.environment:15: 	 	Overriding BSG_IP_CORES_DIR environment variable with Bladerunner defaults.
/home/centos/bsg_bladerunner/bsg_f1/cl_manycore/Makefile.environment:19: 	 	Overriding BSG_MANYCORE_DIR environment variable with Bladerunner defaults.
../../Makefile.common:2: Makefile.deps: No such file or directory
make: *** No rule to make target `Makefile.deps'.  Stop.
```

Is this only compatible with the most up-to-date AMI?

drichmond:
I think you need the corresponding branch in bladerunner. Same name

max-ruttenberg:
OK. On my way to the airport now but I’ll check back on this when I’m stationary.

max-ruttenberg:
@drichmond so I have...

1. cloned `bsg_bladerunner` on F1
2. run `make checkout-all` from `bsg_bladerunner`
3. checked this PR from `bsg_f1`
4. run `make regression`

I'm still seeing this:

```
/home/centos/git/bsg_bladerunner/bsg_f1_9b1ba9e/cl_manycore/Makefile.environment:15: 	 	Overriding BSG_IP_CORES_DIR environment variable with Bladerunner defaults.
/home/centos/git/bsg_bladerunner/bsg_f1_9b1ba9e/cl_manycore/Makefile.environment:19: 	 	Overriding BSG_MANYCORE_DIR environment variable with Bladerunner defaults.
../../Makefile.common:2: Makefile.deps: No such file or directory
make: *** No rule to make target `Makefile.deps'.  Stop.
```

Am I missing a step? It looks to me like the problem is `Makefile.deps` is being included by `Makefile.common` like they are in the same directory (which they are) but the Makefile that is actually including them is two directories deep.

max-ruttenberg:
It looks like the solution to problems like this is using the `MAKEFILE_LIST` variable.

For example, this seems to fix the issue if I replace this line in `Makefile.common` of `bsg_bladerunner`

```include Makefile.deps```

with

```include $(dir $(lastword $(MAKEFILE_LIST)))/Makefile.deps```

Note that ```$(lastword $(MAKEFILE_LIST))``` gives you the absolute path of the current Makefile (not the current working directory).

max-ruttenberg:
OK! I now see that I was just air because I hadn't read your last comment carefully.

I was missing the step where I checkout `feature_bladerunner_vars_defaults` in `bsg_bladerunner`.

Things seem to run smoothly when I do that.

max-ruttenberg:
@drichmond I think this PR is OK but I there's an issue with the related branch in bsg_bladerunner.

PR #234, opened 2019-05-31
New hb_mc_symbol_to_eva function to new API

max-ruttenberg:
* hb_mc_symbol_to_eva() is like symbol_to_eva() but takes a buffer as input instead of a file path.


drichmond:
Merge?

Issue #177, opened 2019-05-03
Add timeouts to functions with while(1) loops


mrutt92:
Functions that can spin indefinitely should have a timeout option.

drichmond:
How should we handle cosimulation?

max-ruttenberg:
That's a good question. Is there no way to count cycles that have passed in COSIM?

drichmond:
There is, now. It's called `sv_bsg_realtime` - see bsg_pr_dbg

PR #237, opened 2019-05-31
Credits fix

leonardxiang:
- The manycore link hardware now checks the receive fifo credits. If the credits are exhausted, host request write will fail.
- Fix the mismatch between the credits counter and the receive fifo elements.

### DIFF
--- a/cl_manycore/regression/library/test_manycore_credits.c
+++ b/cl_manycore/regression/library/test_manycore_credits.c
@@ -49,7 +49,7 @@ int test_manycore_credits() {
 	/* Perform an EVA write that fills DMEM of a single tile.     */
 	/**************************************************************/
 	target = hb_mc_coordinate(x, y);
-	eva = 0x0000;
+	eva = 0x1000;
 
 	for (i = 0; i < dmem_words; ++i){
 		write_data[i] = rand() % ((1 << 16)-1);
@@ -61,7 +61,7 @@ int test_manycore_credits() {
 			__func__, eva, x, y);
 		goto cleanup;
 	}
-
+	bsg_pr_info("Successfully fill DMEM of tile (%d, %d)\n", x, y);
 	rc = hb_mc_manycore_eva_read(mc, &default_map, &target, &eva, read_data, dmem_size);
 				
 	if (rc != HB_MC_SUCCESS) {
--- a/hdl/axil_to_mcl.v
+++ b/hdl/axil_to_mcl.v
@@ -39,6 +39,7 @@ module axil_to_mcl
   , data_width_p="inv"
   , max_out_credits_p="inv"
   , load_id_width_p="inv"
+  , rcv_fifo_els_lp=256
   , link_sif_width_lp=`bsg_manycore_link_sif_width(addr_width_p,data_width_p,x_cord_width_p,y_cord_width_p,load_id_width_p)
   , packet_width_lp=`bsg_manycore_packet_width(addr_width_p,data_width_p,x_cord_width_p,y_cord_width_p,load_id_width_p)
   , return_packet_width_lp=`bsg_manycore_return_packet_width(x_cord_width_p,y_cord_width_p,data_width_p,load_id_width_p)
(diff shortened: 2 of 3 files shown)


cl_manycore/libraries/bsg_manycore.cpp
```diff
@@ -802,7 +802,8 @@ int hb_mc_manycore_packet_tx(hb_mc_manycore_t *mc,
 		}
 	}
 
-	do { //	wait until transmit is complete: continuously write a packet length until done
+	int trials = 10;
```
max-ruttenberg: Is there a reason why the magic number here is 10?

leonardxiang: There is no special reason for that. It is a maximum number of write tails to make the regression test fail.

drichmond: So the purpose of this added while loop is to satisfy a regression test?

leonardxiang: Yes, only for the test_manycore_credits.c.
Where the credits are used up, then the host should not write anymore.

drichmond: But the host shouldn't care about the number of credits - the hardware should.

@max-ruttenberg Is this related to #177 ?

drichmond: @leonardxiang The host should check if the FIFO is full

leonardxiang: In this case, the Tx FIFO not full, but the host's credits is used up, so the host interface will not accept the read request sending from the host.
Do you mean that the host should check the credits is not 0 before sending the read request?

drichmond: But the purpose of this Issue/PR is to make the AXI-Manycore interface manages credits. When there are no credits available, packets should not be dequeued from the hardware FIFO. 

The host should only check if the Hardware FIFO is full.

drichmond: The host should never have to check that the credits are non-zero.

leonardxiang: Agreed.

drichmond: So the purpose of this check is merely to return a fail to the regression test after a certain amount of time?

leonardxiang: Yes.

drichmond: And if these lines are not present, what happens in the regression test?

leonardxiang: It will hang forever

drichmond: OK. It will hang forever because the packet was never sent. 

@max-ruttenberg, same question, #177 would address this correct?

max-ruttenberg: Yes it would. There’s a timeout argument already, we just need to use it.

drichmond: So should we address it in this PR, or in another?

leonardxiang: The RX part should also be fixed.

https://github.com/bespoke-silicon-group/bsg_f1/blob/0be751214ce3375a22fca8b228b3873a5e17301b/cl_manycore/libraries/bsg_manycore.cpp#L867-L876

I think they should be addressed in another PR.

drichmond: OK, can you update your PR to remove the while loop?

The test passes without this while loop, right?

drichmond: I meant, remove the trials while loop. We still need to check for tx_complete

drichmond: Sorry. I'll try to be clearer on this. 

The while loop was there previously right? Can you return this file to its unmodified state? I don't believe any software needs to change for this PR

drichmond:
Awesome. Thanks for the quick turn around on this one!

It passes in cosim right?

drichmond:
Leonard, can you make this a PR to master? we're no longer using dev

PR #238, opened 2019-05-31
Master fixes loader off by one

max-ruttenberg:


drichmond:
This has some extra commits in it now. Can you rebase and eliminate those?

max-ruttenberg:
I don't know how to do that.

PR #239, opened 2019-05-31
Fixed issue where appended file paths caused a comment in C

drichmond:
* Called $(abspath ...) before defining BSG_MANYCORE in C Macro


drichmond:
Passes in F1

PR #241, opened 2019-05-31
move host to (0,0)

leonardxiang:
Move host io to origin coordinate and remove the bladerunner rom in NPA address.

This comes with bsg_manycore PR #30


drichmond:
Did this pass in Cosim?

leonardxiang:
It passes all except for the test_manycore_credits which is addressed in another PR.

Message is as follow:

BSG INFO: test_manycore_credits Regression Test (COSIMULATION)
ERROR:   default_eva_to_epa_tile: Invalid EVA Address 0x0. Does not map to an addressible tile memory locatiion.
ERROR:   hb_mc_manycore_eva_write: Error, requested copy to region that is smaller than buffer
ERROR:   test_manycore_credits: failed to write buffer to EVA 0x0 of coord (1,1)
BSG REGRESSION TEST FAILED
BSG COSIM FAIL: Test failed with exit code:          -1
                    0
$finish called at time : 7740 ns : File "/home/xlelephant/lx_git_repos/bsg_f1/cl_manycore/testbenches/cosim/cosim_wrapper.sv" Line 37
exit

drichmond:
Is there a corresponding change in bsg_manycore that moves the host interface to (0,0)?

leonardxiang:
It passed the simulation, But some test cases will send finish packet to coordinate (0,3):

remote_load_crc
bsg_transpose
fib
crc32
quicksort
nprimes
depend_stall_mispredict

I believe that some separate parameters are not set correctly in those tests. But I did not find them.

drichmond:
Did you run make clean? As far as I can tell, all of these call bsg_finish, which uses IO_X_INDEX. You've changed that definition.

PR #247, opened 2019-06-04
Modified occupancy method to account for FIFO width

drichmond:


cl_manycore/libraries/bsg_manycore.cpp
```diff
@@ -132,16 +132,26 @@ static int hb_mc_manycore_rx_fifo_get_occupancy(hb_mc_manycore_t *mc,
...
         int err;
 
-        err = hb_mc_manycore_mmio_read32(mc, occupancy_addr, occupancy);
+        err = hb_mc_manycore_mmio_read32(mc, occupancy_addr, &val);
         if (err != HB_MC_SUCCESS) {
                 manycore_pr_err(mc, "Failed to get %s occupancy\n", typestr);
                 return err;
         }
 
+        // All packets recieved packets should have an integral occupancy that                                                                                                                                                                                                                                                                                                                                                                  // is determined by the Packet Bit-Width / FIFO Bit-Width
```
max-ruttenberg: This comment feels incomplete?

drichmond: Yep

drichmond: Fixed. Linewrap issue

drichmond:
Can't - this function is static

max-ruttenberg:
Fair enough. Do the regression tests break?

drichmond:
@leonardxiang says test_manycore_credits does. I'm checking on something

drichmond:
It currently does not fail on any tests at the moment. The only reason it would fail at the moment is if fifo drain is called with stale data from a previous iteration.

If we had the ability to receive multiple packets at once, that method would fail without this PR.

But the data has been validated on F1

drichmond:
test_manycore_credits fails until batching is fixed

Issue #175, opened 2019-05-03
Parse CUDA Library constants from .elf file


drichmond:
https://github.com/bespoke-silicon-group/bsg_f1/blob/b27a138286b1840dee57de4086bbf5fcfc0d9cea/cl_manycore/libraries/bsg_manycore_cuda.cpp#L15-L21

Is it possible to pull these constants from symbols in the elf file?

drichmond:
DMEM_BASE should also be parsed from there (I think)

drichmond:
Is this stale? (or will be with @bornaehsani's changes?)

bornaehsani:
Not stale, but it is has priority on my todo list. Will try to address it soon.

max-ruttenberg:
This looks like you could do this with symbol_to_eva?

bornaehsani:
Yeah the problem is not on the host side, I can use symbol to eva IF we had a symbol to begin with, problem is I can't define them as a symbols and use them in the riscv assembly, the assembly doesn't take symbols (or at least I couldn't find a way to do it).

vb000:
@bornaehsani Could you refer `crt.S` in bsg_manycore? It uses symbols in assembly...

max-ruttenberg:
@bornaehsani checkout the input/output operands https://gcc.gnu.org/onlinedocs/gcc/Extended-Asm.html

Here's an example.
```c

uint32_t r_global = 0xDEADBEEF;
uint32_t w_global;

int main()
{       
        uint32_t r, w = 0xCAFEBABE;

        // r = r_global
        // w_global = w
        asm ("lw   %0, %2\n"
             "sw   %3, %1\n"
             : "=r"(r), "=m"(w_global)
             : "m"(r_global), "r" (w)
             : "memory" );
               
        bsg_printf("r=%u, w_global = %u\n", r, w_global);
               
}
```

bornaehsani:
wow this looks great! I'll try to implement it, thank you both!

bornaehsani:
Addressed in PR #298 .

PR #251, opened 2019-06-05
Merge Refactored CUDA API 

bornaehsani:
This pull request merges the refactored CUDA API into F1 master. 

Summary of modifications: 

1. Added bsg_manycore_cuda.cpp/h into the cl_manycore f1 libraries. This file includes all host primitives and routines responsible for
     - Initializing a device with a pool of tiles
     - Loading the binary on all tiles
     - Allocating grids of tile groups on the network
     - Allocating memory on device DRAM, and moving memory to and from device DRAM
     - Sending packets to tiles setting their __bsg_x/y, __bsg_id, __bsg_tile_group_id_x/y, __bsg_grp_org_x/y, CSR_TGO_X/Y, __bsg_grid_dim_x/y symbols. 
     - Sending packets to tiles setting the argc, argv, finish signal address and kernel pointer. 
     - Scheduling grids of tile groups to launch and execute on the device
     - Synchronizing with tile groups to wait for their finish packet

2. Separated the spmd and cuda regression test directories. Added the following regression tests for cuda. Each test has a corresponding directory in the dev_cuda_v6 (to be master) branch of the bsg_manycore/software/spmd/bsg_cuda_lite_runtime. 
     - test_empty: Launches a 1x1 grid of 2x2 tile group to run the empty kernel. Designed to test tile group allocation. 
     - test_empty_parallel: Launches a 4x2 grid of 2x2 tile groups to all run the empty kernel. Designed to test the CUDA API tile group scheduler.
     - test_barrier: Launches a 1x1 grid of 2x2 tile group to run the barrier kernel. Designed to test the functionality of bsg_tile_group_barrier primitive. 
     - test_packet: Launches a 1x1 grid of 2x2 tile group to run the packet kernel. Each tile sends packets with all the symbols that the CUDA API is charged with setting, and the host API picks them up to  validate their value. Designed to test the functionality of sending packets from tiles, and hb_mc_manycore_packet/request_rx routines. 
     - test_stack_load: Launches a 1x1 grid of 2x2 tile group to load 16 arguments (8 go through stack) and store the sum into DRAM. The host then picks up the sum to compare and validate stack usability. Designed to test the functionality of loading arguments from stack for kernels. 
     - test_dram: Launches a gird of 2x2 tile groups to store specific values into DRAM that are then picked up and analyzed by the host interface. Designed to test the functionality of hb_mc_malloc and hb_mc_memcpy routines. 
     - test_shared_mem: Launches a 1x1 grid of 2x2 tile group to run the shared_mem kernel. Tiles in a  tile group declare a tile group shared memory, store to it and load with different striping algorithms to evaluate the function of barrier and shared memory. Designed to test the functionality of bsg_tile_group_shared_mem primitive. 
     - test_shared_mem_load_store: Launches a grid of multiple tile groups to run the shared_mem_load_store kernel. Host loads a MxN matrix into device DRAM, each tile loads a portion into tile group shared memory and stores to another location. Host the picks up the data and compares. Designed to test the functionality of tile group shared memory in conjunction with DRAM load/stores. 
     - test_vec_add: Launches a 1x1 grid of 2x2 tile group to run vec_add kernel. Host loads two arrays into device dram and picks up the result to compare.
     - test_vec_add_parallel: Launches a grid of multiple 2x2 tile groups to perform the vec_add_parallel kernel. Each tile group handles a portion of the vector addition.
     - test_vec_add_parallel_multi_grid: Launches two different grids of 2x2 tile groups in parallel to perform vector addition on two different vectors separately.
     - test_vec_add_parallel_multi_grid: Launches two different grids of 2x2 tile groups sequentially to perform vector addition on two different vectors separately.
     - test_vec_add_shared_mem: Launches a grid of multiple 2x2 tile groups to perform the vec_add_parallel kernel using tile group shared memory. Each tile group handles a portion of the vector addition.
     - test_matrix_mul: Launches a grid of multiple 2x2 tile groups to perform matrix_mul kernel. Each tile group handles a portion  (block_size_y * block_size_x) of the result matrix.
     - test_matrix_mul_shared_mem: Launches a grid of multiple 2x2 tile groups to perform matrix_mul with shared memory. Not working.
All cuda regression tests are heavily documented and include commentary for each section of the code.

Regression tests passed for all library/spmd/cuda regression tests. Rebased, no conflict, and ready to merge.


cl_manycore/libraries/bsg_manycore_cuda.cpp
```diff
@@ -30,11 +36,14 @@ static int hb_mc_get_x(tile_t *tiles, uint8_t *x_list, uint32_t num_tiles) {
 		return HB_MC_FAIL;
 	}
 	for (int i = 0; i < num_tiles; i++) {
-		x_list[i] = tiles[i].x;
+		x_list[i] = tiles[i].coord.x;
```
max-ruttenberg: Please use the setter functions for `hb_mc_coordinate_t` -- hb_mc_coordinate_t is currently a struct but someday it might be a 64-bit integer and then code using the `.` syntax would break.

cl_manycore/libraries/bsg_manycore_cuda.cpp
```diff
@@ -47,312 +56,858 @@ static int hb_mc_get_y(tile_t *tiles, uint8_t *y_list, uint32_t num_tiles) {
 		return HB_MC_FAIL;
 	}
 	for (int i = 0; i < num_tiles; i++) {
-		y_list[i] = tiles[i].y;
+		y_list[i] = tiles[i].coord.y;
```
max-ruttenberg: See https://github.com/bespoke-silicon-group/bsg_f1/pull/251/files#r290994726

cl_manycore/libraries/bsg_manycore_cuda.cpp
```diff
@@ -47,312 +56,858 @@ static int hb_mc_get_y(tile_t *tiles, uint8_t *y_list, uint32_t num_tiles) {
...
-	if (symbol_to_eva(elf, "_bsg_dram_end_addr", &program_end) != HB_MC_SUCCESS) {
+int hb_mc_mesh_init (device_t *device, hb_mc_dimension_t dim){ 
+
+	int error;
+
+	const hb_mc_config_t *cfg = hb_mc_manycore_get_config(device->mc);
+	hb_mc_dimension_t device_dim = cfg->dimension;
+
+	
+	if (dim.x <= 0){
```
max-ruttenberg: See https://github.com/bespoke-silicon-group/bsg_f1/pull/251/files#r290994726

cl_manycore/libraries/bsg_manycore_cuda.cpp
```diff
@@ -47,312 +56,858 @@ static int hb_mc_get_y(tile_t *tiles, uint8_t *y_list, uint32_t num_tiles) {
...
-int hb_mc_device_init (uint8_t *fd, eva_id_t eva_id, char *elf, tile_t *tiles, uint32_t num_tiles) {
-	
-	int error = hb_mc_fifo_init(fd); 
-	if (error != HB_MC_SUCCESS) {
-		fprintf(stderr, "hb_mc_device_init() --> hb_mc_init_host(): failed to initialize host.\n");
+int hb_mc_tile_group_allocate_tiles (device_t *device, tile_group_t *tg){
+	int error;
+	if (tg->dim.x > device->mesh->dim.x){
+		bsg_pr_err("%s: tile group X dimension (%d) larger than mesh X dimension (%d).\n", __func__, tg->dim.x, device->mesh->dim.x);
 		return HB_MC_FAIL;
```
max-ruttenberg: You should check `bsg_manycore_errno.h` for more expressive error codes. `HB_MC_INVALID` might be a better choice here.

cl_manycore/libraries/bsg_manycore_cuda.cpp
```diff
@@ -47,312 +56,858 @@ static int hb_mc_get_y(tile_t *tiles, uint8_t *y_list, uint32_t num_tiles) {
...
-	if (eva_id != 0) {
-		return HB_MC_FAIL; /* eva_id not supported */
-	} 
-	
-	for (int i = 0; i < num_tiles; i++) { /* initialize tiles */
-		hb_mc_tile_freeze_dep(*fd, tiles[i].x, tiles[i].y);
-		hb_mc_tile_set_group_origin(*fd, tiles[i].x, tiles[i].y, tiles[i].origin_x, tiles[i].origin_y);
+	if (tg->dim.y > device->mesh->dim.y){
+		bsg_pr_err("%s: tile group Y dimension (%d) larger than mesh Y dimension (%d).\n", __func__, tg->dim.y, device->mesh->dim.y);
+		return HB_MC_FAIL;
```
max-ruttenberg: see https://github.com/bespoke-silicon-group/bsg_f1/pull/251/files#r290995539

cl_manycore/libraries/bsg_manycore_cuda.cpp
```diff
@@ -47,312 +56,858 @@ static int hb_mc_get_y(tile_t *tiles, uint8_t *y_list, uint32_t num_tiles) {
...
 	}
+	for (hb_mc_idx_t org_y = device->mesh->origin.y; org_y <= (device->mesh->origin.y + device->mesh->dim.y - tg->dim.y); org_y++){
+		for (hb_mc_idx_t org_x = device->mesh->origin.x; org_x <= (device->mesh->origin.x + device->mesh->dim.x - tg->dim.x); org_x++){
+			int free = 1;
+			hb_mc_idx_t tile_id;
+			for (hb_mc_idx_t x = org_x; x < org_x + tg->dim.x; x++){
+				for (hb_mc_idx_t y = org_y; y < org_y + tg->dim.y; y++){
+					tile_id = (y - device->mesh->origin.y) * device->mesh->dim.x + (x - device->mesh->origin.x);
```
max-ruttenberg: I recommend making lines of code like this a function. It looks like you do this exact thing in a few different places. You'll thank yourself later.

cl_manycore/libraries/bsg_manycore_cuda.cpp
```diff
@@ -47,312 +56,858 @@ static int hb_mc_get_y(tile_t *tiles, uint8_t *y_list, uint32_t num_tiles) {
...
-	} 
-	
-	for (int i = 0; i < num_tiles; i++) { /* initialize tiles */
-		hb_mc_tile_freeze_dep(*fd, tiles[i].x, tiles[i].y);
-		hb_mc_tile_set_group_origin(*fd, tiles[i].x, tiles[i].y, tiles[i].origin_x, tiles[i].origin_y);
+	if (tg->dim.y > device->mesh->dim.y){
+		bsg_pr_err("%s: tile group Y dimension (%d) larger than mesh Y dimension (%d).\n", __func__, tg->dim.y, device->mesh->dim.y);
+		return HB_MC_FAIL;
 	}
+	for (hb_mc_idx_t org_y = device->mesh->origin.y; org_y <= (device->mesh->origin.y + device->mesh->dim.y - tg->dim.y); org_y++){
```
max-ruttenberg: What exactly is this looping over? I wan't to say the mesh but I'm honestly not sure.

cl_manycore/libraries/bsg_manycore_cuda.cpp
```diff
@@ -47,312 +56,858 @@ static int hb_mc_get_y(tile_t *tiles, uint8_t *y_list, uint32_t num_tiles) {
...
+ * @param[in]  device        Pointer to device
+ * @param[in]  tg            Pointer to tile group
  * @return HB_MC_SUCCESS on success and HB_MC_FAIL on failure. 
  */
-int hb_mc_device_init (uint8_t *fd, eva_id_t eva_id, char *elf, tile_t *tiles, uint32_t num_tiles) {
-	
-	int error = hb_mc_fifo_init(fd); 
-	if (error != HB_MC_SUCCESS) {
-		fprintf(stderr, "hb_mc_device_init() --> hb_mc_init_host(): failed to initialize host.\n");
+int hb_mc_tile_group_allocate_tiles (device_t *device, tile_group_t *tg){
```
max-ruttenberg: I recommend reworking this function (just a little):
1. It has a couple repeated one-liners that you should consider making their own function.
2. It has several more levels of control flow than a single function should have.
3. Some comments would be very helpful to future you (and to others).

cl_manycore/libraries/bsg_manycore_cuda.cpp
```diff
@@ -47,312 +56,858 @@ static int hb_mc_get_y(tile_t *tiles, uint8_t *y_list, uint32_t num_tiles) {
...
+		if (device->tile_groups == NULL) {
+			bsg_pr_err("%s: failed to allocate space for tile_group_t structs.\n", __func__);
+			return HB_MC_FAIL;
+		}
+	}
 	
-	for (int i = 0; i < num_tiles; i++) { /* freeze tiles */
-		hb_mc_tile_freeze_dep(fd, tiles[i].x, tiles[i].y);
+
+	tile_group_t* tg = (tile_group_t *)(device->tile_groups + device->num_tile_groups) ;
```
max-ruttenberg: I recommend avoiding pointer arithmetic. The above line of code is equivalent to this (which is less confusing):
```c 
tile_group_t* tg = &device->tile_groups[device->num_tiles_groups]; 
```

cl_manycore/libraries/bsg_manycore_cuda.cpp
```diff
@@ -47,312 +56,858 @@ static int hb_mc_get_y(tile_t *tiles, uint8_t *y_list, uint32_t num_tiles) {
...
-		fprintf(stderr, "hb_mc_device_finish() --> hb_mc_host_finish(): failed to terminate host.\n");
+	tg->kernel = (kernel_t *) malloc (sizeof(kernel_t));
+	if (tg->kernel == NULL) { 
+		bsg_pr_err("%s: failed to allocated space for kernel_t struct.\n", __func__);
 		return HB_MC_FAIL;
 	}
+	tg->kernel->name = name;
+	tg->kernel->argc = argc;
+	tg->kernel->argv = argv;
+	tg->kernel->finish_signal_addr = FINISH_BASE_ADDR + ((tg->id.y * tg->grid_dim.x + tg->id.x) << 2); /* TODO: Hardcoded */
```
max-ruttenberg: I recommend making this a function. Maybe something like `tile_get_finish_signal_addr()`?

cl_manycore/libraries/bsg_manycore_cuda.cpp
```diff
@@ -47,312 +56,858 @@ static int hb_mc_get_y(tile_t *tiles, uint8_t *y_list, uint32_t num_tiles) {
...
-		return HB_MC_FAIL; /* invalid EVA ID */
-
-	else if (kind == hb_mc_memcpy_to_device) { /* copy to Manycore */
-		eva_t dst_eva = (eva_t) reinterpret_cast<uintptr_t>(dst);
-		for (int i = 0; i < count; i += sizeof(uint32_t)) { /* copy one word at a time */
-			char *src_word = (char *) src + i;
-			int error = hb_mc_cpy_to_eva(fd, eva_id, dst_eva + i, (uint32_t *) (src_word)); 		
-			if (error != HB_MC_SUCCESS)
-				return HB_MC_FAIL; /* copy failed */
+	device->program->bin_name = strdup (bin_name);
```
max-ruttenberg: `strdup` can return NULL.

cl_manycore/libraries/bsg_manycore_cuda.cpp
bornaehsani: Done, thanks!

cl_manycore/libraries/bsg_manycore_cuda.cpp
bornaehsani: Done.

cl_manycore/libraries/bsg_manycore_cuda.cpp
bornaehsani: Good catch, fixed!

cl_manycore/libraries/bsg_manycore_cuda.cpp
bornaehsani: Yes basically these outer loops are iterating over the entire mesh to find the origin of the new tile group we want to allocate, inside there are another two loops that check to see if all the tiles in a tile group starting from this origin are free, I'll add some comments on this, I agree it is confusing.

cl_manycore/libraries/bsg_manycore_cuda.cpp
bornaehsani: Brilliant! fixed, thanks!

cl_manycore/libraries/bsg_manycore_cuda.cpp
max-ruttenberg: OK cool. Thank you!

cl_manycore/libraries/bsg_manycore_cuda.cpp
```diff
@@ -47,312 +56,860 @@ static int hb_mc_get_y(tile_t *tiles, uint8_t *y_list, uint32_t num_tiles) {
...
+
+
+						error = hb_mc_tile_set_grid_dim_symbols(device->mc, tg->map, device->program->bin, device->program->bin_size, &(device->mesh->tiles[tile_id].coord), &(tg->grid_dim));
+						if (error != HB_MC_SUCCESS) { 
+							bsg_pr_err("%s: failed to set tile (%d,%d) grid size symbol __bsg_grid_size.\n", __func__, hb_mc_coordinate_get_x(device->mesh->tiles[tile_id].coord), hb_mc_coordinate_get_y(device->mesh->tiles[tile_id].coord));
+							return HB_MC_FAIL;
+						}
+					}
+				}
+
```
max-ruttenberg:
```c
tg->origin = hb_mc_coordinate(org_x,org_y)
```

cl_manycore/libraries/bsg_manycore_cuda.cpp
```diff
@@ -47,312 +56,860 @@ static int hb_mc_get_y(tile_t *tiles, uint8_t *y_list, uint32_t num_tiles) {
...
+	// Iterate over the entire mesh as tile (org_y, org_x) being the origin of the new tile group to allcoate 
+	for (hb_mc_idx_t org_y = hb_mc_coordinate_get_y(device->mesh->origin); org_y <= (hb_mc_coordinate_get_y(device->mesh->origin) + hb_mc_dimension_get_y(device->mesh->dim) - hb_mc_dimension_get_y(tg->dim)); org_y++){
+		for (hb_mc_idx_t org_x = hb_mc_coordinate_get_x(device->mesh->origin); org_x <= (hb_mc_coordinate_get_x(device->mesh->origin) + hb_mc_dimension_get_x(device->mesh->dim) - hb_mc_dimension_get_x(tg->dim)); org_x++){
+			int free = 1;
+			hb_mc_idx_t tile_id;
+			
+			// Iterate over a tg->dim.x * tg->dim.y square of tiles starting from the (org_x,org_y) checking to see if all tiles are free so they can be allocated to tile group
+			for (hb_mc_idx_t x = org_x; x < org_x + hb_mc_dimension_get_x(tg->dim); x++){
+				for (hb_mc_idx_t y = org_y; y < org_y + hb_mc_dimension_get_y(tg->dim); y++){
+					tile_id = (y - hb_mc_coordinate_get_y(device->mesh->origin)) * hb_mc_dimension_get_x(device->mesh->dim) + (x - hb_mc_coordinate_get_x(device->mesh->origin));
```
max-ruttenberg: Please make a function called something like this:
```c
hb_mc_idx_t get_tile_id(mesh, coordinate)
```
and replace instances of this computation with that function.

cl_manycore/libraries/bsg_manycore_cuda.cpp
```diff
@@ -47,312 +56,860 @@ static int hb_mc_get_y(tile_t *tiles, uint8_t *y_list, uint32_t num_tiles) {
...
 	}
 
+	// Iterate over the entire mesh as tile (org_y, org_x) being the origin of the new tile group to allcoate 
+	for (hb_mc_idx_t org_y = hb_mc_coordinate_get_y(device->mesh->origin); org_y <= (hb_mc_coordinate_get_y(device->mesh->origin) + hb_mc_dimension_get_y(device->mesh->dim) - hb_mc_dimension_get_y(tg->dim)); org_y++){
+		for (hb_mc_idx_t org_x = hb_mc_coordinate_get_x(device->mesh->origin); org_x <= (hb_mc_coordinate_get_x(device->mesh->origin) + hb_mc_dimension_get_x(device->mesh->dim) - hb_mc_dimension_get_x(tg->dim)); org_x++){
+			int free = 1;
+			hb_mc_idx_t tile_id;
+			
+			// Iterate over a tg->dim.x * tg->dim.y square of tiles starting from the (org_x,org_y) checking to see if all tiles are free so they can be allocated to tile group
+			for (hb_mc_idx_t x = org_x; x < org_x + hb_mc_dimension_get_x(tg->dim); x++){
```
max-ruttenberg: This double for-loop can moved into a function like this one:
```c
bool tile_group_is_free(meh, org_x, org_y)
```

cl_manycore/libraries/bsg_manycore_cuda.cpp
```diff
@@ -47,312 +56,860 @@ static int hb_mc_get_y(tile_t *tiles, uint8_t *y_list, uint32_t num_tiles) {
...
+			}
+			if (free){ // tg->dim.x * tg->dim.y group of free tiles are found at origin (org_x, org_y)
+				hb_mc_coordinate_t org = hb_mc_coordinate(org_x, org_y); 
+				error = hb_mc_origin_eva_map_init (tg->map, org); 
+				if (error != HB_MC_SUCCESS) { 
+					bsg_pr_err("%s: failed to initialize grid %d tile group (%d,%d) eva map origin.\n", __func__, tg->grid_id, hb_mc_coordinate_get_x(tg->id), hb_mc_coordinate_get_y(tg->id));
+					return HB_MC_FAIL;
+				}
+				
+				for (hb_mc_idx_t x = org_x; x < org_x + hb_mc_dimension_get_x(tg->dim); x++){
```
max-ruttenberg: Can this double for-loop be moved into a function like this?
```c
int allocate_tile_group(device, tg)
```

cl_manycore/libraries/bsg_manycore_cuda.cpp
max-ruttenberg: please make it `static` -- nobody should be calling but you.

cl_manycore/libraries/bsg_manycore_cuda.cpp
max-ruttenberg: please make it `static` -- nobody should be calling but you.

cl_manycore/libraries/bsg_manycore_cuda.cpp
max-ruttenberg: please make it `static` -- nobody should be calling but you.

cl_manycore/libraries/bsg_manycore_cuda.cpp
drichmond: Max's point is good. To expand on what he's saying (motivate, perhaps) is that people feel that more than three levels of indentation in code is detrimental to reading code. 

FWIW, this is codified in the Linux Kernel Style Guide: 

https://www.kernel.org/doc/html/v4.10/process/coding-style.html#indentation


Something, something, self documenting code... https://en.wikipedia.org/wiki/Self-documenting_code

drichmond:
Borna, are you comfortable squashing the "minor fix" commits?

cl_manycore/libraries/bsg_manycore_cuda.cpp
max-ruttenberg: That is more or less my point for some of my other comments. In this comment I’m referring to how ‘tile_id’ is computed.  I guess what it’s doing is finding the index of a tile in some management vector maintained by the mesh? I count five arithmetic ops (which I would say makes this a “complicated” line of code) and this exact line of code seems to occur at multiple lines in this file. That’s why I recommend making it a function. Tell the compiler you want it in-lined if it makes you feel better (but I really wouldn’t worry about that).

cl_manycore/libraries/bsg_manycore_cuda.cpp
```diff
@@ -47,312 +56,858 @@ static int hb_mc_get_y(tile_t *tiles, uint8_t *y_list, uint32_t num_tiles) {
...
+			tiles[tile_id].origin.x = default_origin.x;
+			tiles[tile_id].origin.y = default_origin.y;
+			tiles[tile_id].tile_group_id.x = -1;
+			tiles[tile_id].tile_group_id.y = -1;
+			tiles[tile_id].free = 1;
+		}
+	}
+
+	mesh_t *mesh = new mesh_t; 
+	mesh->dim.x = dim.x;
```
max-ruttenberg: Please use the setters and getters here too please.

cl_manycore/libraries/bsg_manycore_cuda.cpp
drichmond: Fair, I saw the for loops and conflated it with some of your other comments. 

Oh to be in Python and just have an iterator...

cl_manycore/libraries/bsg_manycore_cuda.h
```diff
@@ -15,27 +15,289 @@
...
+
+
+/**
+ * Initializes a program's memory allcoator and creates a memroy manager
+ * @param[in]  program       Pointer to program
+ * @param[in]  id            Id of program's memory allocator
+ * @param[in]  name    Unique name of program's memory allocator
+ * @return HB_MC_SUCCESS on success and HB_MC_FAIL on failure.
+ */
+int hb_mc_program_allocator_init (hb_mc_program_t *allocator, const char *name, hb_mc_allocator_id_t id);
```
max-ruttenberg: Why does the caller need to pass the allocator name here?

cl_manycore/libraries/bsg_manycore_cuda.h
```diff
@@ -15,27 +15,289 @@
...
+__attribute__((warn_unused_result))
+int hb_mc_grid_init (device_t *device, hb_mc_dimension_t grid_dim, hb_mc_dimension_t tg_dim, char *name, uint32_t argc, uint32_t argv[]);
+
+
+
+
+/**
+ * Searches for a free tile group inside the device mesh and allocoates it, and sets the dimensions, origin, and id of tile group.
+ * @param[in]  device        Pointer to device
+ * @param[in]  tg            Pointer to tile group
```
max-ruttenberg: Is `tg` an input or an output?

cl_manycore/regression/cuda/test_dram.c
```diff
@@ -0,0 +1,157 @@
...
+	if (rc != HB_MC_SUCCESS) { 
+		bsg_pr_err("failed to de-initialize device.\n");
+		return HB_MC_FAIL;
+	}
+
+
+	int mismatch = 0; 
+	for (int i = 0; i < N; i++) {
+		if (A_host[i] != i) { 
+			bsg_pr_err(BSG_RED("Mismatch") ": -- A[%d] = 0x%x\t Expected: 0x%x\n", i , A_host[i], i);
```
max-ruttenberg: I recommend formatting 32-bit hex numbers like this: `"%08" PRIx32`
1. More important: The `"08"` pads with 0s in the upper digits and the resulting print statements, I find, are much easier to read.
2. Less important, but good to know: The `PRIx32` is defined in `<inttypes.h>` and it addresses the fact that `%x` is for the type of size `unsigned int` or `int` but that does not mean a 32-bit integer on every system.

cl_manycore/libraries/bsg_manycore_cuda.cpp
bornaehsani: Done, thanks!

cl_manycore/libraries/bsg_manycore_cuda.cpp
bornaehsani: Fixed, great point!

cl_manycore/libraries/bsg_manycore_cuda.cpp
bornaehsani: Well this was stupid of me! Fixed, thanks!

max-ruttenberg: Don't think that! It had to think about it too.

cl_manycore/libraries/bsg_manycore_cuda.cpp
bornaehsani: Yeah that makes sense, I should thank you! I added the following, what do you think? 
hb_mc_get_relative_coordinate (origin, coord): takes an absolute coordinate and an origin and subtract to return the relative coord. 
hb_mc_coordinate_to_flat_index (coord, dim): converts 2D coord to 1D index
hb_mc_dimension_to_size (dim) : converts 2D dimension to 1D size.

drichmond: What do flat_index and _size mean? 

Max and I typically reserve size for things referring to bytes.

cl_manycore/libraries/bsg_manycore_cuda.cpp
drichmond: Max's feedback here is good too. Just to emphasize the point - I spent several days on EVA translation, and you only saw the end of it. We iterated on it for a long time until it was committed, and then even more after. 

One example is the comments/methods in eva to npa DRAM translation. They were originally in one method, with block comments, then they were split into separate methods with block comments just to make it easier to maintain.

cl_manycore/libraries/bsg_manycore_cuda.cpp
max-ruttenberg: I think what you know mean by "flat index" (it's a 2D array flattened to 1D) but you can just call it an index -- after all that's what it is.
@drichmond is right -- we like to use `size` to mean "size in bytes" like the `sizeof` operator in C. Maybe just use `length`?
So is the plan to implement `hb_mc_coordinate_to_index()` in terms of `hb_mc_dimension_to_length()` and `hb_mc_get_relative_coordinate()`?

cl_manycore/libraries/bsg_manycore_cuda.cpp
bornaehsani: Agreed, it was way too complicated. So I broke it down into three functions: 
1. hb_mc_tile_group_allocate_tiles: the entire function
2. hb_mc_tiles_are_free: to do the job of the inner two loops, searches starting from an origin to see if all tiles within the tile group dimensions radius are free. 
3. hb_mc_group_initialize_tiles: called after we found a group of free tiles. It sends packets to set all the symbols in the tiles. 
Let me know what you think!

cl_manycore/libraries/bsg_manycore_cuda.cpp
bornaehsani: This all makes perfect sense. To that end, I created a static hb_mc_get_tile_id routine and replaced all instances with this routine in the code. 
I also broke the hb_mc_tile_group_allocate_tiles into three separate routines called hb_mc_tile_group_allocate_tiles, hb_mc_tiles_are_free, and hb_mc_tile_group_initialize_tiles to be more readable. Let me know what you think!

cl_manycore/libraries/bsg_manycore_cuda.cpp
bornaehsani: Yes agreed, moved the double for loop into hb_mc_tiles_are_free(device, origin, dim).

cl_manycore/libraries/bsg_manycore_cuda.cpp
bornaehsani: Agreed, moved it to hb_mc_tile_group_initialize_tiles(device, tg). Will make it static.

cl_manycore/libraries/bsg_manycore_cuda.cpp
bornaehsani: Thanks for spotting it! Fixed.

cl_manycore/libraries/bsg_manycore_cuda.h
bornaehsani: It is an input. The function tries to find a group of free tiles for this specific tile group, and once it's found it sends packets to set the tile symbols using the data inside tg struct.

cl_manycore/regression/cuda/test_dram.c
bornaehsani: interesting, very nice pointer! I replaced all 0x%x notations with this one, thanks!

cl_manycore/libraries/bsg_manycore_cuda.cpp
bornaehsani: ha :D

cl_manycore/libraries/bsg_manycore_cuda.cpp
bornaehsani: Wow I should hit refresh every once in a while, just saw your replies! 
@drichmond, so as Max said, the flat_index is the conversion of 2D coordinates to 1D index, and you're right I will just call it index. And I will also change size to length.
@max-ruttenberg  actually I implemented it the other way around, so 'hb_mc_coordinate_to_index' just flattens a coordinate, and 'hb_mc_dimension_to_length' had nothing to do with flattening, I just built it to replace anywhere in code where is use dim.x * dim.y to get the total number of tiles. 
I added another function as you suggested called 'hb_mc_get_tile_id (origin, dim, coord) that uses 'hb_mc_get_relative_coordinate' and 'hb_mc_coordinate_index' to get the final tile id. Does that sound good?

bornaehsani:
> 
> 
> Borna, are you comfortable squashing the "minor fix" commits?

@drichmond  I squashed the minor fix, clean up and minor cleanup commits. Thanks for pointing it out!

bornaehsani:
@max-ruttenberg @drichmond I believe all issues are addressed. I'd love your final review on this so we can move forward. Thanks a lot to both of you!

cl_manycore/libraries/bsg_manycore_coordinate.h
```diff
@@ -19,7 +23,7 @@ typedef enum __hb_mc_coordinate_limit_t {
         HB_MC_COORDINATE_MAX = 32
 } hb_mc_coordinate_limit_t;
 
-typedef int32_t hb_mc_idx_t;
+typedef uint32_t hb_mc_idx_t;
```
drichmond: Is there any reason you changed this to a uint32_t?

cl_manycore/libraries/bsg_manycore_cuda.cpp
```diff
@@ -5,354 +5,1089 @@
...
+#include <bsg_manycore.h>
+#include <bsg_manycore_printing.h>
+#include <bsg_manycore_eva.h>
+#include <bsg_manycore_origin_eva_map.h>
 
-static const uint32_t KERNEL_REG = 0x1000 >> 2; //!< EPA of kernel. 
-static const uint32_t ARGC_REG = 0x1004 >> 2; //!< EPA of number of arguments kernel will use. 
-static const uint32_t ARGV_REG = 0x1008 >> 2; //!< EPA of arguments for kernel. 
-static const uint32_t SIGNAL_REG = 0x100c >> 2; //!< EPA of register that holds signal address. Tile will write to this address once it completes the kernel.   
+static const hb_mc_epa_t KERNEL_REG = 0x1000; //!< EPA of kernel.
```
drichmond: Can you define these as Macros instead of static const variables?

cl_manycore/libraries/bsg_manycore_cuda.cpp
```diff
@@ -5,354 +5,1089 @@
...
+static const hb_mc_epa_t ARGV_REG   = 0x1008; //!< EPA of arguments for kernel. 
+static const hb_mc_epa_t SIGNAL_REG = 0x100c; //!< EPA of register that holds signal address. Tile will write to this address once it completes the kernel.   
 
 
-static const uint32_t FINISH_ADDRESS = 0xC0DA; //!< EPA to which tile group sends a finish packet once it finishes executing a kernel  
-
-static awsbwhal::MemoryManager *mem_manager[1] = {(awsbwhal::MemoryManager *) 0}; /* This array has an element for every EVA <-> NPA mapping. Currently, only one mapping is supported. */
+static const hb_mc_epa_t FINISH_BASE_ADDR = 0xF000; //!< EPA to which tile group sends a finish packet once it finishes executing a kernel  
 
 static uint32_t const DRAM_SIZE = 0x80000000;
```
drichmond: I believe there is a method for getting the DRAM size.

cl_manycore/libraries/bsg_manycore_cuda.cpp
drichmond: And can you make the naming scheme similar to other defined addresses? See mmio.h, where I think the standard is HB_MC_MMIO_<NAME>_ADDR

drichmond:
Hi Borna, have you had a chance to run this on one of the newer bladerunner images, like v0.6.0, or ideally, v0.6.2?

cl_manycore/libraries/bsg_manycore_cuda.h
```diff
@@ -15,27 +15,289 @@
...
+
+
+/**
+ * Initializes a program's memory allcoator and creates a memroy manager
+ * @param[in]  program       Pointer to program
+ * @param[in]  id            Id of program's memory allocator
+ * @param[in]  name    Unique name of program's memory allocator
+ * @return HB_MC_SUCCESS on success and HB_MC_FAIL on failure.
+ */
+int hb_mc_program_allocator_init (hb_mc_program_t *allocator, const char *name, hb_mc_allocator_id_t id);
```
max-ruttenberg: `__attribute__((warn_unused_result))`

max-ruttenberg:
Would you mind making a best effort at restricting code line size to 100 characters?
Long lines of code can be challenging to read on smaller monitors.
After you've done that and addressed @drichmond's outstanding comments, I think we're ready to merge this in.

cl_manycore/libraries/bsg_manycore_coordinate.h
bornaehsani: Well first of all I figured why have it as int32 since a negative index doens't make sense, and secondly int the previous API before the refactor all indexes were declared as uint32, so I figured it would be safer to match that. Is there any reason why having it has int32 would make more sense?

max-ruttenberg: I'm not sure why it was a signed int to begin with. @drichmond why was this made signed?

drichmond:
I had them as int32 because things will REALLY fail if a negative number is used.

It doesn’t look like we check for a valid target when we send the packet though. I’ll file an issue for that and we can leave this as is.

In general though, you shouldn’t be editing things that aren’t relevant to the issue you’re addressing. It’s better to file a separate PR so that changes don’t sneak into the codebase

max-ruttenberg:
Why wouldn't the packet API check that the coordinate is in the correct range? It has the dimensions of the fabric from the configuration.

cl_manycore/libraries/bsg_manycore_driver.h
```diff
@@ -144,9 +144,13 @@ int hb_mc_all_host_req_complete(uint8_t fd);
 int hb_mc_get_config(uint8_t fd, hb_mc_config_id_t id, uint32_t *cfg);
 uint8_t hb_mc_get_manycore_dimension_x();
 uint8_t hb_mc_get_manycore_dimension_y();
+uint8_t hb_mc_get_host_intf_coord_x();
```
drichmond: These methods are provided by the config API. Is there a reason you didn't use those?

cl_manycore/libraries/bsg_manycore_driver.h
```diff
@@ -144,9 +144,13 @@ int hb_mc_all_host_req_complete(uint8_t fd);
...
+uint8_t hb_mc_get_host_intf_coord_y();
 
 int hb_mc_fifo_finish(uint8_t fd);
+
```
drichmond: Was it necessary to add this whitespace?

cl_manycore/libraries/bsg_manycore_loader.cpp
```diff
@@ -946,3 +946,40 @@ int hb_mc_loader_symbol_to_eva(const void *bin, size_t sz, const char *symbol,
 
         return HB_MC_SUCCESS;
 }
+
+
+int hb_mc_loader_read_program_file(const char *file_name, unsigned char **file_data, size_t *file_size)
```
drichmond: Can you add a comment describing the purpose of this function?

cl_manycore/libraries/bsg_manycore_loader.h
```diff
@@ -37,6 +37,10 @@ int hb_mc_loader_load(const void *bin, size_t sz,
 int hb_mc_loader_symbol_to_eva(const void *bin, size_t sz, const char *symbol,
                                hb_mc_eva_t *eva);
 
+
+int hb_mc_loader_read_program_file(const char *file_name, unsigned char **file_data, size_t *file_size);
```
drichmond: Can you add a comment describing the purpose of this function?

cl_manycore/libraries/bsg_manycore_mem.cpp
```diff
@@ -121,8 +121,10 @@ static int hb_mc_npa_is_dram (npa_t *npa) {
  * checks if NPA is in host endpoint.
  */
 static int hb_mc_npa_is_host (npa_t *npa) {
-	uint8_t dim_x = hb_mc_get_manycore_dimension_x();
-	if ((npa->y == 0) && (npa->x == dim_x))
+	uint8_t coord_x = hb_mc_get_host_intf_coord_x();
```
drichmond: Can you use the hb_mc_idx_t type instead of uint8_t?

cl_manycore/libraries/bsg_manycore_mem.h
```diff
@@ -3,6 +3,7 @@
 
 #include "bsg_manycore_features.h"
```
drichmond: Can you use angle brackets instead of "?

cl_manycore/libraries/bsg_manycore_origin_eva_map.cpp
```diff
@@ -0,0 +1,84 @@
+#include <bsg_manycore_origin_eva_map.h>
+#include <bsg_manycore_eva.h>
+#include <bsg_manycore_printing.h>
+#include <cstring>
```
drichmond: Can you use the ifdef __cplusplus macro to switch between c headers and c++ headers?

cl_manycore/libraries/bsg_manycore_tile.cpp
```diff
@@ -81,102 +82,408 @@ int hb_mc_tile_set_origin(hb_mc_manycore_t *mc, const hb_mc_coordinate_t *tile,
 	return rc;
 }
 
+
```
drichmond: Was this additional whitespace necessary?

cl_manycore/libraries/bsg_manycore_tile.cpp
```diff
@@ -1,6 +1,7 @@
 #include <bsg_manycore_tile.h> 
 #include <bsg_manycore_driver.h>
 #include <bsg_manycore_epa.h>
+#include <stdio.h>
```
drichmond: Can you use the C++ macro to switch between C++ headers and C headers?

cl_manycore/libraries/bsg_manycore_tile.cpp
```diff
@@ -81,102 +82,408 @@ int hb_mc_tile_set_origin(hb_mc_manycore_t *mc, const hb_mc_coordinate_t *tile,
...
+/*! 
+ * Writes a uint8_t into a tile given its coordinates and epa 
+ * Behavior is undefined if #mc is not initialized with hb_mc_manycore_init().
+ * @param[in] mc         A manycore instance initialized with hb_mc_manycore_init().
+ * @param[in] coord      Cooridnates of the destination tile
+ * @param[in] epa        Epa address in desitnation tile to which the write occurs
+ * @param[in] v          Variable to be written into tile 
+ * @return HB_MC_SUCCESS on success and HB_MC_FAIL on failure.
+* */
+int hb_mc_tile_write8(hb_mc_manycore_t *mc, const hb_mc_coordinate_t *coord, const hb_mc_epa_t *epa, uint8_t v) {
```
drichmond: Can you use the hb_mc_manycore_write API directly instead of creating a new API?

cl_manycore/regression/cuda/Makefile
```diff
@@ -0,0 +1,12 @@
+include Makefile.tests
+include ../../Makefile.dimensions
+include ../Makefile.common
+
+$(TARGETS): test_%: $(BSG_MANYCORE_DIR)/software/spmd/bsg_cuda_lite_runtime/%/binary
+
+$(BSG_MANYCORE_DIR)/software/spmd/bsg_cuda_lite_runtime/%/binary:
+	$(MAKE) -C $(dir $@) clean main.riscv \
+		bsg_tiles_X=2 \
```
drichmond: Can you use the dimensions parsed from Makefile.deps like we do in the spmd tests?

https://github.com/bespoke-silicon-group/bsg_f1/blob/dbf65db99b9032df52baa7debe6cbd0b2d3c2d22/cl_manycore/regression/spmd/Makefile#L9

cl_manycore/regression/cuda/cuda_tests.h
```diff
@@ -0,0 +1,31 @@
+#ifndef __CUDA_TESTS_H
+#define __CUDA_TESTS_H
+
+#ifdef COSIM
```
drichmond: Can you remove the ifdef COSIM and use the angle bracket includes for everything except the simulation libraries?

cl_manycore/regression/spmd/spmd_tests.h
```diff
@@ -18,6 +18,7 @@
 #include <bsg_manycore_mem.h>
 #include <bsg_manycore_loader.h>
 #include <bsg_manycore_errno.h>
+
```
drichmond: Whitespace change?

drichmond:
@max-ruttenberg Do we check valid coordinates in the packet API? I didn't see a check, so I filed an issue

cl_manycore/libraries/bsg_manycore_cuda.cpp
bornaehsani: Good point! I converted them to macro defines, changed the name in accordance with the naming schemes, and moved them to the header file. 
The plan is to remove them entirely and load these values from the binary file (issue #175 ), I looked into this a bit but its a bit complicated, since these values are being used in the riscv assembly and I still haven't found a way to integrate defined macros or variables into the assembly. I will address this issue in a separate pull request.

drichmond: We can look into it! Maybe I don't understand the purpose.

The most future proof would be to define a getter function that takes a binary and returns the EVA (?) and the calling EVA to NPA to get the EPA. 

But the macros are good for now.

cl_manycore/libraries/bsg_manycore_coordinate.h
drichmond: Because I thought that we could return some codes using negative integers.

cl_manycore/libraries/bsg_manycore_cuda.cpp
```diff
@@ -5,354 +5,1089 @@
...
+			device->mesh->tiles[tile_id].tile_group_id = tg->id;
+			device->mesh->tiles[tile_id].free = 0;
+
+			hb_mc_coordinate_t coord_val = hb_mc_coordinate((x - hb_mc_coordinate_get_x(origin)), (y - hb_mc_coordinate_get_y(origin)));
+
+
+			error = hb_mc_tile_set_origin_registers(device->mc, &(device->mesh->tiles[tile_id].coord), &(device->mesh->tiles[tile_id].origin));
+			if (error != HB_MC_SUCCESS) { 
+				 bsg_pr_err("%s: failed to set tile (%d,%d) tile group origin registers CSR_TGO_X/Y.\n", __func__, hb_mc_coordinate_get_x(device->mesh->tiles[tile_id].coord), hb_mc_coordinate_get_y(device->mesh->tiles[tile_id].coord));
+				return HB_MC_FAIL;
```
drichmond: You might want to consider returning the value of error, instead of HB_MC_FAIL.

max-ruttenberg:
@drichmond no. But if we did, I would check that it's less than the dimension (not that it's positive). Of course if it's signed, then I also need to check that it's gte 0.

drichmond:
Sounds good to me.

cl_manycore/libraries/bsg_manycore_coordinate.h
max-ruttenberg: Why would we return error codes where we should have returned a coordinate value? If a function can return an error, it should be returning an int.

cl_manycore/libraries/bsg_manycore_cuda.cpp
bornaehsani: Yes you are right, thanks for pointing it out! I removed the parameter and used the hb_mc_config_get_dram_size() instead.

cl_manycore/libraries/bsg_manycore_cuda.h
bornaehsani: Great catch! fixed, thanks!

cl_manycore/libraries/bsg_manycore_cuda.h
```diff
@@ -21,7 +21,7 @@ extern "C" {
 #define HB_MC_MMIO_TILE_ARGV_PTR_EPA		0x1008	// EPA of pointer to argument list of kernel in tile.
```
drichmond: How about HB_MC_CUDA_TILE_ARGV_PTR_EPA

This is my fault. I meant you to look at the naming scheme there to demonstrate how we've named similar macros, not to copy them verbatim. MMIO is used for addresses that are accessible by AXI. 

I would also recommend adding them to the base offset defined in tile. For example: 

```
define HB_MC_MMIO_CUDA_ARGV_PTR_EPA (HB_MC_TILE_EPA_DMEM_BASE  + 0x08)

drichmond: This means that if we change the EPA base, then these constants will automatically change as well

cl_manycore/libraries/bsg_manycore_coordinate.h
drichmond: Relax, I didn't say it was a well thought out idea

bornaehsani:
> 
> 
> Hi Borna, have you had a chance to run this on one of the newer bladerunner images, like v0.6.0, or ideally, v0.6.2?

@drichmond yes I just ran them on the 6.2, everything passes, I'll also double check all tests on the newest images before merging once all comments are addressed.

drichmond:
Great to hear! Thanks Borna!

I don't anticipate any further updates at the moment. Maybe v0.6.3 for the PRs I made this morning.

bornaehsani:
Sure! the v0.6.3 is not up on the bladerunner yet right? I'll check that one too once the image is made.

drichmond:
No, because my PR’s haven’t been approved yet

bornaehsani:
> 
> 
> Would you mind making a best effort at restricting code line size to 100 characters?
> Long lines of code can be challenging to read on smaller monitors.
> After you've done that and addressed @drichmond's outstanding comments, I think we're ready to merge this in.
Hey @max-ruttenberg , yes that makes sense to me. Are there any specific guidelines I should follow in doing this? For example is breaking a for loop or a function call into more than one line OK?

drichmond:
For loops are odd to break into multiple lines, but I’m fine with it. If you need to, you can define your bounds before the for-loop instantiation.

Definitely break functions onto multiple lines if you can.

cl_manycore/libraries/bsg_manycore_driver.h
bornaehsani: They are actually deprecated, I've already replaced them with the methods in config API. Should I delete them or leave them with a depreacted tag? (leaving them with a deprecated tag generates a lot of warnings on compile, too messy in my opinion)

bornaehsani:
OK sounds good, will do!

cl_manycore/libraries/bsg_manycore_driver.h
bornaehsani: ha! happened by mistake :D I added a deprecated tag to each one, but it generated 20 warnings on compile so I removed them but forgot to remove the whitespace. This brings me to a philosophical question, is it better to remove the deprecated functions all together or leave them at the cost of generating 100 warnings? They only reason they were left there in the first place was to make sure the cuda API doesn't break all at once, but now that we don't need them anymore should I remove them?

cl_manycore/libraries/bsg_manycore_driver.h
drichmond: Deprecated tag.

drichmond:
Let’s do a separate PR after. 100s of messages are fine

cl_manycore/libraries/bsg_manycore_loader.cpp
bornaehsani: My bad, forgot. Fixed, thanks!

cl_manycore/libraries/bsg_manycore_mem.cpp
bornaehsani: These functions are deprecated, which is why I've left them with the deprecated get_host_intf_coord() function (it returns a uint8_t so I didn't change the type). I'll add deprecated tags to all of them.

cl_manycore/libraries/bsg_manycore_cuda.cpp
bornaehsani: Yes the problem isn't on the host side I can handle that, the problem is that I have to define these variables on the device code so they show up in the elf as symbols, and to use the variables in the assembly code on the device side (I can't integrate variables or even a define macro into riscv assembly code, I've been using magic numbers for now), that's the thing I have to figure out.

cl_manycore/libraries/bsg_manycore_mem.cpp
drichmond: Right on

cl_manycore/libraries/bsg_manycore_cuda.cpp
max-ruttenberg: yeah that sounds good.

cl_manycore/libraries/bsg_manycore_driver.h
drichmond: Deprecated tags. Then we also know what is deprecated

cl_manycore/libraries/bsg_manycore_tile.cpp
bornaehsani: Done!

cl_manycore/libraries/bsg_manycore_cuda.cpp
bornaehsani: Great pointer! I replaced them in the library and the regression tests. thanks!

cl_manycore/libraries/bsg_manycore_cuda.h
bornaehsani: Ahh I see, yeah that makes sense! Done, thanks!

cl_manycore/libraries/bsg_manycore_tile.cpp
bornaehsani: guess not! fixed!

cl_manycore/regression/spmd/spmd_tests.h
bornaehsani: Well it wasn't hurting anybody but I fixed it anyway :D

cl_manycore/regression/cuda/cuda_tests.h
bornaehsani: Nice catch. Fixed it for all three cuda, library and spmd headers.

cl_manycore/regression/cuda/Makefile
bornaehsani: @drichmond So basically I had to change two things:
1. In the Makefile I set the dimensions to CL_MANYCORE_DIM_X/Y to be used for building the hardware with these  dimensions
2. I added TILE_GROUP_DIM_X/Y to the Makefile.dimensions, which are used in Makefile.test for compiling the riscv binary. Because the riscv dimensions should not be same as the entire mesh dimensions anymore, they should match the tile group dimensions otherwise barrier won't work, as Max found out a while ago. 
What do you think?

bornaehsani:
@drichmond @max-ruttenberg , I believe all issues are addressed, I also tried my best to reduce code line sizes, and I ran all regression tests on the latest image v0.6.2 all passed. Let me know what you think, thanks a lot!

drichmond:
Sounds reasonable to me

drichmond:
I think this is awesome! @max-ruttenberg Are you ready to ship it?

max-ruttenberg:
Looks good

bornaehsani:
Thanks a lot to both of you!

PR #254, opened 2019-06-07
Refactored a few constants into config

drichmond:
* Moved DMEM SIZE from Tile to Config
* Added config parameter to local EVA translation methods
* Unified DRAM Size Calculation


drichmond:
Passes on F1 v0.6.2

(@bornaehsani, there are no API changes here, but you should check that you didn't use any of the constants I removed)

bornaehsani:
@drichmond thanks for letting me know! It doesn't seem to have any conflict with the cuda PR (I havedn't used the deleted constant).

PR #263, opened 2019-06-09
Master responder

max-ruttenberg:
* adds an API for having responders that automatically respond to manycore request packets
* adds a spmd loader executable


cl_manycore/executables/bsg_manycore_spmd_loader.cpp
```diff
@@ -0,0 +1,75 @@
+#include "bsg_manycore_run_spmd_program.hpp"
+#include <bsg_manycore.h>
+#include <bsg_manycore_printing.h>
+#include <stdlib.h>
+#include <stdio.h>
```
bornaehsani: Would it be useful to separate cpp headers and c headers with ifdef?

cl_manycore/libraries/bsg_manycore_request_packet.h
```diff
@@ -0,0 +1,307 @@
...
+#define BSG_MANYCORE_REQUEST_PACKET_H
+
+#include <bsg_manycore_features.h>
+#include <bsg_manycore_bits.h>
+#include <bsg_manycore_errno.h>
+#include <bsg_manycore_epa.h>
+#include <endian.h>
+#include <inttypes.h>
+#include <stdint.h>
+#include <stdio.h>
```
bornaehsani: Would it be useful to separate cpp headers and c headers with ifdef?

bornaehsani: Would it be useful to separate cpp headers and c headers with ifdef?

cl_manycore/executables/bsg_manycore_spmd_loader.cpp
drichmond: Quick learner :p

max-ruttenberg: I hate to be spicy about this but... the only difference it makes (that I know of) is that the C library functions are imported as part of the `std` namespace -- which we really don't use when calling C library functions because... why would we? It's borderline sociopathic to make your own function literally called `printf` or `exit`. Worse still, this only applies to standard C library functions -- not to `unistd.h` or `endian.h` or `elf.h` or a whole bunch of other system headers we might want.

So sure, I guess I can change it; but I think we have much better things to worry about than whether or not we can call `printf` like `std::printf`.

bornaehsani: ha :D

Issue #92, opened 2019-04-10
hb_mc_init_device() and device_launch() should accept memory buffer


cylinbao:
https://github.com/bespoke-silicon-group/bsg_f1/blob/b15423845a81f05858d7e4061d47b4707237ff33/cl_manycore/libraries/bsg_manycore_driver.cpp#L432
https://github.com/bespoke-silicon-group/bsg_f1/blob/b15423845a81f05858d7e4061d47b4707237ff33/cl_manycore/libraries/bsg_manycore_driver.cpp#L842
These two functions take a string of a file location and load the binary from the file system.
But in some use cases, the binary will be in the memory already and we just want to pass the memory pointer to the function.
I think it will be better if this function can take memory pointer as input.
For the cases the binary is on the file system, maybe user just write some simple function to read it to the memory and pass it to the function.

[Reference: cuLaunchKernel](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__EXEC.html#group__CUDA__EXEC_1gb8f3dc3031b40da29d5f9a7139e52e15)

drichmond:
@cylinbao Is this resolved in Borna's new API?

cylinbao:
Hi thanks for addressing this.
This isn't resolved directly yet.
As here
https://github.com/bespoke-silicon-group/bsg_f1/blob/e48613ddb11c7b8d622653974e98211414927928/cl_manycore/libraries/bsg_manycore_cuda.cpp#L596
It still get file_path as input.
But it is solved in a lower level api with this
https://github.com/bespoke-silicon-group/bsg_f1/blob/e48613ddb11c7b8d622653974e98211414927928/cl_manycore/libraries/bsg_manycore_cuda.cpp#L638
Maybe @bornaehsani can add a program init function that calls hb_mc_loader_load() more straightly.

bornaehsani:
Hi Chien-Yu! Thanks a lot for pointing this out. Yes this is still not implemented, which is why I have left it open for now. But I will implement it very soon.

bornaehsani:
This issue is now addressed in PR #265 .

PR #265, opened 2019-06-10
Binary loader that accepts buffer instead of path to binary

bornaehsani:
This pull request addresses issue #92 .
Now users can either call hb_mc_device_program_init that takes in path to binary and automatically loads into a buffer, or call hb_mc_device_program_init_binary with a prepared buffer that contains the binary and binary size. 

hb_mc_device_program_init is separated into three functions: 
1. hb_mc_device_program_init(): load binary from file into buffer and call 2
2. hb_mc_device_program_init_binary(): copy binary buffer and set all variables in device's program_t struct and call 3
3. hb_mc_device_program_load(): loads the binary onto all tiles and unfreezes them.

Cuda regression tests passed, ready to merge.

### DIFF
--- a/cl_manycore/libraries/bsg_manycore_cuda.cpp
+++ b/cl_manycore/libraries/bsg_manycore_cuda.cpp
@@ -886,19 +886,25 @@ int hb_mc_device_program_load (hb_mc_device_t *device) {
 
 
 
+
 /**
- * Freezes tiles, loads program binary into all tiles and into dram,
- * and set the symbols and registers for each tile.
+ * Takes in a buffer containing binary and its size,
+ * freezes tiles, loads program binary into all tiles and into dram,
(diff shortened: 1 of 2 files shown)


drichmond:
@cylinbao I'll merge this when you approve it.

drichmond:
@cylinbao 

https://github.com/bespoke-silicon-group/bsg_f1/blob/12617602a16b781418463f639d31abeb1adae9a0/cl_manycore/libraries/bsg_manycore_cuda.cpp#L916

So that it can be tracked and used for printing error messages

drichmond:
It doesn't necessarily need to be a valid path (I believe). Just a name

bornaehsani:
Dustin is right! The name doens't have to be valid, but you need to pass a dummy value for tracking purposes. Thanks to both of you!

PR #297, opened 2019-06-19
Fix the returning data to endpoint

leonardxiang:
Fix the returning data to the manycore.

library and spmd simulation test passed.

basejump_stl: b0374c2
bsg_manycore: 75219ce

### DIFF
--- a/hdl/axil_to_mcl.v
+++ b/hdl/axil_to_mcl.v
@@ -423,15 +423,7 @@ module axil_to_mcl
     // fifo response to manycore
     assign fifo_ready_li[2*i+1] = ~returning_wr_v_r[i];
     assign fifo_res_cast[i]     = fifo_data_lo[2*i+1];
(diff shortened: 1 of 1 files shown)


drichmond:
Hi Leonard, 

Is this a bug you solved? I don't understand what this PR addresses

leonardxiang:
> Hi Leonard,
> 
> Is this a bug you solved? I don't understand what this PR addresses

Hi, Dustin,
This is a bug if the host is working as a slave and response to the manycore request. 
Although currently we don't have this situation in bladerunner, it is good to be fixed.

drichmond:
Cool, good catch. Can you file an issue so that we can track this bug? 

Even if it’s a simple description, reference the line of code so that we can track this in the future.

drichmond:
Addresses #299

leonardxiang:
Address issue #299

Issue #296, opened 2019-06-18
AXI_MEMORY_MODEL is not set to 1 when running from inside testbenches directory


drichmond:
Simple fix

PR #301, opened 2019-06-19
Fixed an issue where AXI_MEMORY_MODEL=0 didn't disable the Axi Memory Model in simulation…

drichmond:
Also addressed #296


max-ruttenberg:
Did this pass COSIM and F1?

drichmond:
Yes, it does.

PR #302, opened 2019-06-20
Hotfix! Fixes bug in hb_mc_manycore_dst_npa_is_valid

bornaehsani:
Fixes a bug in hb_mc_manycore_dst_npa_is_valid()


drichmond:
I don't think there's a bug in npa_valid. NPAs are absolute - they aren't origin based.

Issue #300, opened 2019-06-19
CUDA Lite needs reloading fpga image after each run


bornaehsani:
After each cuda lite test run, the program leaves fpga in a bad state, causing the next run to hang (tiles are unresponsive) 
Multiple tile group executions and kernel executions can run inside one <device_init  -  device_finish> scope, but not inside multiple scopes one after another. 
TODO:
    - create a regression test that reproduces the issue in one program 
    - Test in cosim
    - Fix

sampsyo:
We're running into this too. It would be awesome to be able to run multiple programs without resetting the FPGA!

drichmond:
Yeah, this is a high-priority bug 😬

taylor-bsg:
Do you guys use FPGA style resets in the FPGA code? Is there a dependence
on the FPGA resetting state via bitstream?

Code like:

logic foo=1

Would do that.

Unless the FPGA tools handle this reset case in a way that is not obvious
to me.

drichmond:
Not that I know of.

drichmond:
@leonardxiang can you comment?

drichmond:
It appears that this has been solved in the FPU branch - and it was an issue with load reservation instructions. @max-ruttenberg found the source of the issue, Borna found the behavior in simulation, but by the time we root caused it it had been solved in the FPU branch. 

@tommydcjung Can you comment on what may have been the cause and solution? @max-ruttenberg and @bornaehsani Can chip in as necessary.

drichmond:
Addressed in #325

PR #305, opened 2019-06-20
Added a new cuda regression test to initialize, load binary run and finish twice

bornaehsani:
This regression test is related to issue #300 but doesn't fix it yet.
On both F1 and cosim, the first kernel runs fine but the second kernel hangs.
**Should be merged in with bsg_manycore PR # 67.**

### DIFF
--- a/cl_manycore/regression/cuda/Makefile.tests
+++ b/cl_manycore/regression/cuda/Makefile.tests
@@ -2,10 +2,11 @@ INDEPENDENT_MAIN_TESTS_TYPE = cuda
 
 UNIFIED_MAIN_TESTS = 
 # For F1, only run one test at a time, and reload the agfi image in between runs. To be fixed.
-INDEPENDENT_MAIN_TESTS = test_empty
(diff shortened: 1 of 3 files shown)


drichmond:
Awesome! @soctar If we give you a .vcd file would you be able to put it into the undebugger?

soctar:
Most definitely! If it's from VCS, it's helpful to have the VPD file as well, on the chance that you'd like to decompose structs.

drichmond:
I can give you the vpd file instead!

drichmond:
Or rather - @bornaehsani Can you give Neil the vpd file that is generated when you run this test?

bornaehsani:
Sorry for the late response, and yes of course, however the file seems too large to send over, but Neil and I are planning to meet tomorrow so I can learn how to set up the debugger, perhaps we can go over it tomorrow together. Thanks!

taylor-bsg:
Definitely suggest the smallest binary that manifests the problem, so you
are hunting for your needle in the small haystack possible. :-) VCD files
are very compressible, as well, but you can hand the file off on xor
diskbits.

PR #314, opened 2019-06-26
Bugfix for xdim_max computation in bsg_manycore_config

drichmond:
Found this while trying to increase the manycore dimensions


drichmond:
This is passing in cosim. If someone has an instance up and can test in F1 that would be 🎉

bornaehsani:
@drichmond, Most tests for F1 fail for the v1.0.2 and v3.0.3 with the following error: 
image

The xdim_max is 3 in these cases and the X dimension is 4, I think the xdim_max should be set to (1 << xlogsz_max) instead of (1 << xlogsz_max) - 1

drichmond:
Right! I’ll fix that

bornaehsani:
Great, F1 on v1.0.2: All library/spmd/cuda tests passed
on v3.0.3 All library/spmd/cuda tests passed except test_rom(expected)

PR #322, opened 2019-06-29
added vanilla_core_trace

tommydcjung:


tommydcjung:
we probably need to surround vanilla_core_trace instantiation with something so that this is not included for synthesis

cl_manycore/hardware/cl_manycore.sv
```diff
@@ -665,4 +665,18 @@ cl_debug_bridge CL_DEBUG_BRIDGE (
 );
 
 `endif //  `ifndef DISABLE_VJTAG_DEBUG
+
+
```
drichmond: // synopsys translate on/off will work here - according to the Xilinx manual

PR #332, opened 2019-07-07
Added floating point vector addition test for cuda regression

bornaehsani:
Passes F1.

**Should be merged together with PR # 90 in bsg_manycore.**

### DIFF
--- a/cl_manycore/regression/cuda/test_float_vec_add.c
+++ b/cl_manycore/regression/cuda/test_float_vec_add.c
@@ -1,0 +1,223 @@
+/******************************************************************************/
+/* A[N] + B[N] --> C[N]                                                       */
+/* Runs the floating point vector addition on one 2x2 tile group              */
+/* Grid dimensions are prefixed at 1x1. --> block_size_x is set to N.         */
+/* This tests uses the software/spmd/bsg_cuda_lite_runtime/float_vec_add/     */
+/* manycore binary in the BSG Manycore repository.                            */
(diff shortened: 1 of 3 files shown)


cl_manycore/regression/cuda/test_float_vec_add.c
```diff
@@ -0,0 +1,210 @@
...
+void host_float_vec_add (float *A, float *B, float *C, int N) { 
+	for (int i = 0; i < N; i ++) { 
+		C[i] = A[i] + B[i];
+	}
+	return;
+}
+
+
+int kernel_float_vec_add () {
+	bsg_pr_test_info("Running the CUDA Floating Point Vector Addition Kernel on a 1x1 grid of 2x2 tile group.\n\n");
```
drichmond: Can you line split down to 80 characters?

drichmond: Here's a good reason why. I shouldn't need to scroll sideways to read your code...

image

bornaehsani: Very well, I've modified the test file to limit to 80 characters. Let me know if this looks good, and I'll apply the same to all other tests. Thanks Dustin!

drichmond:
Perfect! 

This is fine for now - no need to redo all of them. We'll fix them when we change them for the no-dram mode (which may become an argument, not a macro)

bornaehsani:
Great! OK will do.

PR #261, opened 2019-06-09
Refactor manycore link using basejump stl

leonardxiang:
Dependency:
BSG_MANYCORE: 8241449
BASEJUMP_STL : e9377ca

PR #341, opened 2019-07-10
Refactored Host Interface to use BSG RTL

drichmond:
This is just a rebased version of #261


drichmond:
Closed. May re-open at a future date.

PR #342, opened 2019-07-10
Support for printing scalars from manycore programs without bsg_printf

max-ruttenberg:
* Supports printing signed 32-bit integers in decimal.
* Supports printing unsigned 32-bit integers in decimal.
* Supports printing unsigned 32-bit integers in hexadecimal.
* Supports printing 32-bit floating point numbers.
* Supports printing 32-bit floating point numbers in scientific notation.
* All scalar values are printed with their source tile coordinate.
* Fixes a compile error in the spmd loader in `executables` directory.


drichmond:
@bornaehsani If you get a chance - can you write a short regression test for print_int?

bornaehsani:
> 
> 
> @bornaehsani If you get a chance - can you write a short regression test for print_int?

@drichmond Addressed in PR #346. Merge that PR before merging this one.

PR #343, opened 2019-07-12
Support for Command Line Arguments to Regression Tests

drichmond:
(Currently WIP)

I've added support for passing command line arguments in cosimulation. This mirrors the argv/arc standard in typical C/C++ programs. In VCS you can now use a "plusargs" argument, c_args, to pass strings that you want to be interpreted as arguments to cosimulation programs

e.g. `./test_rom +c_args="space separated arguments here"` will be parsed and passed as argc/argv inside of cosimulation.

To my knowledge, this is working in cosimulation. But I haven't ported the changes to F1 (yet)


max-ruttenberg:
The parts of this that I understand LGTM. I see that the regression tests call `get_argc()` and `get_argv()` but then don't do much with them? Is there a test that actually used the command line?

drichmond:
See the test_unified_main in spmd and Python.

I haven’t add this in the CUDA tests yet - though I’ve considered passing the path to the RISC-V binary there as well

Dustin

drichmond:
A secondary goal for this PR is to enable all the SPMD tests to be run (@save-buffer), but instead of compiling each as a binary, we now compiled the test_unified_main binary and pass it command line arguments. 

This doesn't work for CUDA tests, since each has a different top-level file.

PR #346, opened 2019-07-19
Regression test for bsg_print_int

bornaehsani:
Related to PR #342 
Adds regression test for bsg_print_int
Also changed deprecated test_packet into test_tile_info that uses bsg_printf

**Should be merged in with PR # 100 on bsg_manycore.**


max-ruttenberg:
At a glance it all looks reasonable. I wonder if you we should also have one that tests this with vanilla SPMD programs? CUDA is complex enough that if one of these tests fails to print it won't be 100% clear if the problem is in CUDA, the manycore side, or the packet responder module. Having an additional vanilla SPMD test could help with this.

Actually testing this feature is a little hard to formulate in the existing infrastructure (wrt tracking if the test failed).

drichmond:
Not sure I understand - but we have a vanilla SPMD test

max-ruttenberg:
@drichmond I don't see it in this PR. These all look like CUDA. Are they already in the repo?

WRT evaluating if the test has passed or failed: what if nothing gets printed but everything else goes right? Does the test fail? I don't think it does and that's my point.

bornaehsani:
@max-ruttenberg @drichmond  You are right, the cuda is too complicated, and as I don't (or shouldn't) have access to the underlying API functions I couldn't check for a fail. I added a SPMD regression test for the scalar printing in PR #347. Let me know what you guys think!

drichmond:
Should this still be merged (with #100 in bsg_manycore?)

PR #364, opened 2019-07-26
Feature cadenv search

drichmond:
* Added support for cadenv

I've added support for cadenv. The default behavior is to use environment variables for VCS and Vivado (since that's necessary for F1/Cornell) but if any UW people want to use cadenv, it will now be found


max-ruttenberg:
This seems reasonable. Tested?

drichmond:
yep

PR #368, opened 2019-07-26
Automatically deduce the environment from inside of BSG Bladerunner

drichmond:
* Added RED to highlight errors
* Reduce warning conditions to actual warnings
* Reworked Makefile.hdk to deduce environment from inside of bladerunner


drichmond:
This passes cosim regression

PR #369, opened 2019-07-30
Makefile Improvements

drichmond:
* Modified clean and PHONY targets in regression and testbenches makfiles


drichmond:
Passes regression

PR #370, opened 2019-07-30
Updated .gitignores

drichmond:


max-ruttenberg:
I approve ignoring `.so` and `.so.1` in the `libraries` directory.

drichmond:
But not ignoring .prelim? 😄

PR #372, opened 2019-07-30
changing instance name so that profiler can match string to print out…

tommydcjung:
… header only once.


tommydcjung:
https://github.com/bespoke-silicon-group/bsg_manycore/blob/40f737306a263937c77042c328feddd8ec985b55/testbenches/common/v/vcache_profiler.v#L73

vcache_profiler needs to check the instance name so that only the first vcache prints out the header.

Issue #349, opened 2019-07-22
Unified regression test for CUDA-Lite regression tests 


bornaehsani:

drichmond:
@save-buffer This might be a useful program for you to get back up to speed

drichmond:
@bornaehsani What CUDA tests are similar enough that they could be run in a single test?

drichmond:
Addressed in #378

PR #378, opened 2019-08-18
Unify four cuda regression tests

save-buffer:
Addresses issue #349


save-buffer:
Specifically, unifies test_empty, test_barrier, test_scalar_print, and test_tile_info

save-buffer:
Passes both COSIM and F1 (though when I ran it on F1 `test_matrix_mul_shared_mem` failed, but I didn't touch that).

PR #380, opened 2019-08-19
Added more ignores to .gitignore

drichmond:


drichmond:
No effect on cosimulation or F1

PR #384, opened 2019-08-20
Switched to using Manycore simulation filelist to reduce duplicated work

drichmond:
This PR reduces the work that Tommy (and myself) need to do everytime the manycore architecture is updated. Instead of copying the manycore file list in F1, we simply use the file list from bsg_manycore

A few related changes are also included


drichmond:
Depends on https://github.com/bespoke-silicon-group/bsg_manycore/pull/132

PR #395, opened 2019-09-10
Fixed logic returning credit packets

vb000:
https://github.com/bespoke-silicon-group/bsg_f1/blob/3af08f4f73c8fc4705205c69fe7d1b3d3492381d/hdl/axil_to_mcl.v#L366

Above logic returns credit packets even the fifo is full. So updated it to return credits only when a in-request is accepted, aka yumi is asserted.

### DIFF
--- a/hdl/axil_to_mcl.v
+++ b/hdl/axil_to_mcl.v
@@ -363,7 +363,7 @@ module axil_to_mcl
   	if(reset_i)
   		returning_wr_v_r <= '0;
   	else
-  		returning_wr_v_r <= endpoint_in_v_lo & endpoint_in_we_lo;
+  		returning_wr_v_r <= endpoint_in_yumi_li & endpoint_in_we_lo;
   end
 
   `declare_bsg_mcl_request_s;


drichmond:
This looks good to me. Thanks for the explanation. @leonardxiang, your thoughts?

leonardxiang:
Looks good to me too.
Thanks a lot! @Bandhav Veluri

drichmond:
Thanks @leonardxiang!

Does the root cause of the issue make sense to you? (i.e. why this is a bug, and how this fix addresses it)

taylor-bsg:
Yes — since this design pattern is very common in our hardware, we want to
make sure you understand why this would fix the problem so that in future
hardware you would be able to avoid it.

M

leonardxiang:
Yes, let me review the bug:

In this issue, the manycore endpoint is master and the manycore link FIFO module is the slave attached.

When the in request from manycore endpoint is remote store, the slave module needs to indicate the commit of the store operation, i.e. by assert the out_response valid signal for 1 cycle.

Here, the handshake of remote store is valid & yumi, so, the store completion is indicated by yumi rather than valid. I handled this incorrectly.


There is also another crack here, according to the **3 rules** for attaching slave module in **BaseJump Manycore Accelerator Network**

https://github.com/bespoke-silicon-group/bsg_f1/blob/3af08f4f73c8fc4705205c69fe7d1b3d3492381d/hdl/axil_to_mcl.v#L432
This line avoids rule 1, the valid & ready signal indicates the commit of slave returning of loaded data, and this signal should delay 1 cycle.

**Rules 1 Response Delay:**  The out_response should be returned at least one cycle after the corresponding in_request. A common case is the store request do not need the returning data and the response can be composed immediately but it is still need to delayed at least one cycle.
**Rules 2 Response Counting:**  The slave should generated an out_response for every consumed in_request. For load request, the out_response is the loaded data; For store request, the loaded data can be any value, but the slave still need to pull up the out_response valid signal for 1 cycle to indicate the commit of the store operation.
**Rules 3 Response Order:**  The out_response must be generated in the order as the in_request.

vb000:
@leonardxiang From my understanding, there are two kinds of responses. Response to a store request, and response to a load request. The why the `|` in code below? https://github.com/bespoke-silicon-group/bsg_f1/blob/3af08f4f73c8fc4705205c69fe7d1b3d3492381d/hdl/axil_to_mcl.v#L432

I think the response delay is fine in this case, because the response is always flopped. But the only issue was that the response should only be sent when the in-request is fully committed to the fifo, which you explained in your explanation.

I haven't looked at the response to the load request though. Are suggesting there's an issue in that as well?

vb000:
And about load-requests, even if fifo output is already valid and ready, that valid output would correspond to a previous load request. So, the response of a request would always be at least on cycle delayed. So, yes, I think this line completely fine in regard to response delay.

https://github.com/bespoke-silicon-group/bsg_f1/blob/3af08f4f73c8fc4705205c69fe7d1b3d3492381d/hdl/axil_to_mcl.v#L432

drichmond:
This will be v3.2.2

drichmond:
@leonardxiang Is everything OK to your knowledge?

PR #407, opened 2019-09-13
Changes to run any SPMD programs from a unified binary -- test_loader.

drichmond:
Simply put: This PR allows users to run SPMD programs using test_loader (formerly test_unified_main). The old program took a test_name, but some users (@mrutt92) would like to re-use this program to load their own out-of-tree SPMD programs. 

I've added this functionality, and then updated the makefiles so that this is unified across simulation and co-simulation. There were a few bumps along the way. 

This had the happy consequence of removing environment variable dependencies from test_loader. I'll remove them from CUDA and Python in subsequent PRs someday.

As written, it passes in Cosimulation and on F1


drichmond:
@mrutt92 this is the second PR

drichmond:
I'm fixing the relative path includes, but other that that, yes it passes in cosim

drichmond:
Passes in cosim 🎉

PR #408, opened 2019-09-16
Added timing to hb_mc_tile_groups_execute() 

bornaehsani:
function to print the total execution time in both cycles and seconds.
Currently timing results seem to be wrong. Do NOT merge until discussed.

### DIFF
--- a/libraries/bsg_manycore_cuda.h
+++ b/libraries/bsg_manycore_cuda.h
@@ -5,8 +5,10 @@
 
 #ifdef __cplusplus
 #include <cstdint>
+#include <ctime>
 #else
 #include <stdint.h>
+#include <time.h>
 #endif
 
 #ifdef __cplusplus
--- a/libraries/bsg_manycore_cuda.cpp
+++ b/libraries/bsg_manycore_cuda.cpp
@@ -1443,6 +1443,7 @@ static int hb_mc_device_wait_for_tile_group_finish_any(hb_mc_device_t *device) {
  */
 int hb_mc_device_tile_groups_execute (hb_mc_device_t *device) {
 
+	clock_t start_time = clock();
 	int error ;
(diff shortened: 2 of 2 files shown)


libraries/bsg_manycore_cuda.h
```diff
@@ -5,8 +5,10 @@
 
 #ifdef __cplusplus
 #include <cstdint>
+#include <ctime>
 #else
 #include <stdint.h>
+#include <time.h>
```
tommydcjung: how about making a stopwatch class like this?
```
#include <chrono>
#include <stdexcept>
#include "stopwatch.h"
using namespace std;
using namespace std::chrono;


/*
 *	default constructor.
 */
Stopwatch::Stopwatch()
{
	_started = false;
}

/*
 * 	start the stopwatch.
 */
void Stopwatch::start()
{
	if (_started)
	{
		throw runtime_error("Stopwatch has already started.");
	}
	_started = true;
	_start = high_resolution_clock::now();
}

/*
 * 	stop the stopwatch and return the duration.
 * 	@return amount of time elapsed in seconds.
 */
double Stopwatch::stop()
{
	if (!_started)
	{
		throw runtime_error("Stopwatch has not started.");		
	}

	duration<double> diff = high_resolution_clock::now() - _start;
	_started = false;	
	return diff.count(); 
}
```

tommydcjung: just an example of an interface.

mrutt92: except we should make it C friendly.

drichmond:
I feel like this might be superseded by the new vanilla stats parser. Is that true?

bornaehsani:
True. Thanks for pointing it out! Addressed in PR # 198 in bsg_manycore.

PR #411, opened 2019-09-17
Branch trace

vb000:
- Added branch trace parameter in Makefile.machines.include
- Added stderr and branch trace epas to uart responder.


libraries/bsg_manycore_uart_responder.cpp
```diff
@@ -3,36 +3,82 @@
 #include <bsg_manycore_printing.h>
 #include <stdio.h>
 
-#define UART_EPA 0xEADC
+enum hb_mc_uart_epa_indx {
+  STDOUT_EPA_INDX, 
+  STDERR_EPA_INDX, 
+  BRANCH_TRACE_EPA_INDX,
```
drichmond: Can you put the branch trace responder code in a separate responder? 

I think STDERR and STDOUT are good in here. But the Branch Trace is performing a fundamentally different task

libraries/bsg_manycore_uart_responder.cpp
```diff
@@ -3,36 +3,82 @@
...
 {
-	auto data = hb_mc_request_packet_get_data(rqst);
-	FILE *f = (FILE*)responder->responder_data;
-	fputc((int)data, f);
+  auto data = hb_mc_request_packet_get_data(rqst);
+  auto src_x = hb_mc_request_packet_get_x_src(rqst);
+  auto src_y = hb_mc_request_packet_get_y_src(rqst);
+
+  for(int i=STDOUT_EPA_INDX; i<HB_MC_NUM_UART_EPAS; i++) {
+    if(hb_mc_request_packet_is_match(rqst, (responder->ids + i))) {
```
mrutt92: we try to avoid pointer arithmetic in this repo. 
note that `responder->ids + i` can be expressed as `&responder->ids[i]`.

libraries/bsg_manycore_uart_responder.cpp
vb000: Isn't it too much work to maintain multiple responders? I think having a generic mmio responder like this makes sense because, the only difference among handling mmios is what we print out to terminal. 
```
      switch (i) {
        case BRANCH_TRACE_EPA_INDX:
          fprintf(f, "BSG_BRANCH_TRACE x=%d y=%d target=%d\n", 
                  src_x, src_y, (int)data);
          break;

        default:
          fputc((int)data, f);
          break;
```
https://github.com/bespoke-silicon-group/bsg_f1/blob/8b121d4df53c69f16a89693a1bbf95bc5048f9fe/libraries/bsg_manycore_uart_responder.cpp#L77

Above switch is the only unique thing for each mmio epa.

libraries/bsg_manycore_uart_responder.cpp
```diff
@@ -3,36 +3,82 @@
...
+
 static int init(hb_mc_responder_t *responder,
-		hb_mc_manycore_t *mc)
+    hb_mc_manycore_t *mc)
 {
-	bsg_pr_dbg("hello from %s\n", __FILE__);
-	responder->responder_data = stdout;
+  bsg_pr_dbg("hello from %s\n", __FILE__);
+
+  for(int i=STDOUT_EPA_INDX; i<HB_MC_NUM_UART_EPAS; i++) {
```
mrutt92: This work can be done at compile time like so:
```c
static hb_mc_request_packet_id_t ids [] = {
    [STDOUT_EPA_INDX]  = RQST_ID( RQST_ID_ANY_X, RQST_ID_ANY_Y, RQST_ID_ADDR(0xEADC),
    [STDERR_EPA_INDX]  = RQST_ID( RQST_ID_ANY_X, RQST_ID_ANY_Y, RQST_ID_ADDR(0xEEE0),
    [BRANCH_TRACE_EPA_INDX] = RQST_ID( RQST_ID_ANY_X, RQST_ID_ANY_Y, RQST_ID_ADDR(0xEEE4), 
};

static FILE * uart_streams [] = {
    [STDOUT_EPA_INDX] = stdout,
    [STDERR_EPA_INDX] = stderr,
    [BRANCH_TRACE_EPA_INDX] = stderr,
}; 
```
personally, I find that more readable than a finite for loop with an inner switch statement.

vb000: Thanks! I used for loop as a way to use enums for ordering, I was not aware of this syntactic sugar, this is lot more elegant!

libraries/bsg_manycore_trace_responder.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+static int respond(hb_mc_responder_t *responder,
+       hb_mc_manycore_t *mc,
+       const hb_mc_request_packet_t *rqst)
+{
+  auto data = hb_mc_request_packet_get_data(rqst);
+  auto src_x = hb_mc_request_packet_get_x_src(rqst);
+  auto src_y = hb_mc_request_packet_get_y_src(rqst);
+
+  for(int i=0; i<HB_MC_NUM_TRACE_EPAS; i++) {
```
mrutt92: please remove unnecessary control flow and please fix the formatting.

libraries/bsg_manycore_uart_responder.cpp
```diff
@@ -3,36 +3,53 @@
...
-		   const hb_mc_request_packet_t *rqst)
+       hb_mc_manycore_t *mc,
+       const hb_mc_request_packet_t *rqst)
 {
-	auto data = hb_mc_request_packet_get_data(rqst);
-	FILE *f = (FILE*)responder->responder_data;
-	fputc((int)data, f);
+  auto data = hb_mc_request_packet_get_data(rqst);
+
+  for(int i=STDOUT_EPA_INDX; i<HB_MC_NUM_UART_EPAS; i++) {
```
mrutt92: see my other comment.

libraries/bsg_manycore_trace_responder.cpp
vb000: Sorry, I'm confused -- don't we need some kind of control flow to detect which responder id in `responder->ids []` the request matches with? Are you suggesting we add a function to do that job?

mrutt92: The for-loop is unnecessary because HB_MC_NUM_TRACE_EPAS=1. Unless there's a compelling reason why that would change the for-loop is just clutter in the code. The if statement is unnecessary because respond() is only called if the packet matches your id table; that pattern matching is handled in the code that calls this callback function.

vb000: Resolved this offline.

PR #414, opened 2019-09-18
Added support for pulling AFI filelist from manycore

drichmond:


vb000:
Cool!

drichmond:
Unfortunately, this doesn't work. Two problems: first, it gets stuck in design synthesis (still trying to debug this). Second, we have some modules that synthesize incorrectly (bsg_mem_1rw_sync_mask_write_bit.v) or not at all (infinite memories)

drichmond:
This seems to be working now. Removing/adding Xilinx synthesizable files is a bit of a hack, but I added a comment. And it's only one file at the moment.

PR #417, opened 2019-09-19
License

drichmond:


drichmond:
I think I met my weekly quota of LoC changes

Issue #337, opened 2019-07-10
Freeing of bin can cause a double free if CUDA initialized with program_init_binary()


mrutt92:
Suppose CUDA is initialized with:
```c
hb_mc_device_progam_init_binary(hb_mc_device_t *device, 
                                     const char *bin_name,
                                     const unsigned char* bin_data, 
                                     size_t bin_size, 
                                     const char* alloc_name, 
                                     hb_mc_allocator_id_t id)
```
This line of code will cause an invalid free: the consequences of this are dire.

https://github.com/bespoke-silicon-group/bsg_f1/blob/39ae1eb895d0bdc3010dca6d8821546d8e5ff244/cl_manycore/libraries/bsg_manycore_cuda.cpp#L1236

drichmond:
Was this ever addressed?

bornaehsani:
> 
> 
> Was this ever addressed?

Oh god I totally forgot about this. I submitted a PR for the fix, plus a regression test for it.

bornaehsani:
Addressed in PR #419.

PR #419, opened 2019-09-22
Fixes the double freeing of binary bug

bornaehsani:
Related to issue #337.
- Added static hb_mc_program_binary_copy() CUDA API function to copy the binary buffer into hb_mc_program_t struct instead of pointing to the buffer.
- Added a cuda regression test test_binary_load_buffer to test the correctness of hb_mc_program_init_binary function that loads binary onto tiles from a pre-loaded buffer instead of loading it from a given path, and to test if the double freeing of binary bug is fixed.
F1 passed.
**merge with PR # 166 in bsg_manycore.**

### DIFF
--- a/libraries/bsg_manycore_cuda.cpp
+++ b/libraries/bsg_manycore_cuda.cpp
@@ -115,6 +115,11 @@ __attribute__((warn_unused_result))
 static int hb_mc_device_program_exit (hb_mc_program_t *program); 
 
 __attribute__((warn_unused_result))
+static int hb_mc_program_binary_copy (hb_mc_program_t *program,
+                                      const unsigned char *bin_data,
+                                      size_t bin_size); 
+
+__attribute__((warn_unused_result))
 static int hb_mc_program_allocator_init (const hb_mc_config_t *cfg,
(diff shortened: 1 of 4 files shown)


libraries/bsg_manycore_cuda.cpp
```diff
@@ -1282,6 +1292,34 @@ static int hb_mc_device_program_exit (hb_mc_program_t *program) {
...
+static int hb_mc_program_binary_copy (hb_mc_program_t *program,
+                                      const unsigned char *bin_data,
+                                      size_t bin_size) { 
+	unsigned char* copy = (unsigned char*) malloc (bin_size);
+	if (!copy){ 
+		bsg_pr_err("%s: failed to allocated space on device for program binary.\n", __func__);
+		return HB_MC_NOMEM;
+	}
+
+	for (int ptr = 0; ptr < bin_size; ptr++){
```
mrutt92: use memcpy()

bornaehsani: done! thanks for pointing it out!

PR #425, opened 2019-09-30
Added bsg_tile_group_id flat tile group id symbol 

bornaehsani:
To the bsg_manycore_cuda.cpp source file, to be set and sent to each tile along with bsg_tile_group_id_x and y. 
**Merge with PR # 169 in bsg_manycore.**

### DIFF
--- a/libraries/bsg_manycore_cuda.cpp
+++ b/libraries/bsg_manycore_cuda.cpp
@@ -1996,6 +1996,9 @@ static int hb_mc_device_tiles_set_config_symbols (hb_mc_device_t *device,
 
 
 		// Set tile's index __bsg_x/y symbols.
+		// A tile's __bsg_x/y symbols represent its X/Y 
+		// coordinates with respect to the origin tile 
+		// in that specific tile group
 		hb_mc_idx_t coord_x = hb_mc_coordinate_get_x (coord); 
 		hb_mc_idx_t coord_y = hb_mc_coordinate_get_y (coord); 
 		error = hb_mc_tile_set_symbol_val(device->mc,
@@ -2032,8 +2035,13 @@ static int hb_mc_device_tiles_set_config_symbols (hb_mc_device_t *device,
 
 
 
-		// Set tile's id __bsg_id symbol.
-		// calculate tile's id 
+		// Set tile's __bsg_id symbol.
+		// bsg_id uniquely identifies each tile in a tile group
+		// Tile group is a 2D array of tiles that concurrently run a kernel inside the binary
+		// __bsg_x/y represent the X/Y coordiantes of tile relative to tile group origin
+		// The flat tile id is calculated using tile group dimensions 
+		// and the tile group X/Y coordiantes relative to tile group origin as follows:
+		// __bsg_id = __bsg_y * __bsg_tile_group_dim_x + __bsg_x
 		hb_mc_idx_t id = hb_mc_coordinate_get_y(coord) * hb_mc_dimension_get_x(tg_dim) + hb_mc_coordinate_get_x(coord); 
 		error = hb_mc_tile_set_symbol_val(device->mc,
                                                   map,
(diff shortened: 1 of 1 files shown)


libraries/bsg_manycore_cuda.cpp
```diff
@@ -2053,9 +2053,11 @@ static int hb_mc_device_tiles_set_config_symbols (hb_mc_device_t *device,
 
 
 
-		// Set tile's tile group index  __bsg_tile_group_id_x/y symbol.
+		// Set tile's tile group index  __bsg_tile_group_id_x/y symbols.
+		// and tile's flat tile group id (tg_id_y * grid_dim_x + tg_id_x) 
 		hb_mc_idx_t tg_id_x = hb_mc_coordinate_get_x (tg_id); 
 		hb_mc_idx_t tg_id_y = hb_mc_coordinate_get_y (tg_id);
+                hb_mc_idx_t tg_id = tg_id_y * hb_mc_dimension_get_x(grid_dim) + tg_id_x;
```
drichmond: Can you fix the tabs vs spaces?

bornaehsani:
> 
> 
> This looks great, but why is it necessary?

I need the flat tile group id to use as tag in bsg_print_stats instructions. 
As to why do it on the host, I either have to do the same calculation in each manycore code the determine the flat tile group id, or I can do it once in host and send packets to set this symbol in the tiles. I figured this way it is defined as part of the host overhead, and not the actual kernel execution. Does that make sense?

drichmond:
yeah, it makes sense. But can you add a comment explaining why it's necessary and how it differs from tg_id_x and tg_id_y?

bornaehsani:
> 
> 
> yeah, it makes sense. But can you add a comment explaining why it's necessary and how it differs from tg_id_x and tg_id_y?

yeah that's a good idea, will do, thanks!

libraries/bsg_manycore_cuda.cpp
bornaehsani: thanks for pointing it out, since all lines in this file use tab, I changed this one to tab as well.

bornaehsani:
how does it look now?

libraries/bsg_manycore_cuda.cpp
```diff
@@ -2053,9 +2053,20 @@ static int hb_mc_device_tiles_set_config_symbols (hb_mc_device_t *device,
 
 
 
-		// Set tile's tile group index  __bsg_tile_group_id_x/y symbol.
+		// Set tile's tile group index  __bsg_tile_group_id_x/y symbols.
+		// and tile's flat tile group id (tg_id_y * grid_dim_x + tg_id_x)
+		// __bsg_tile_group_id_x/y symbols define the X/Y coordinates of
+		// a tile's tile group within it's father grid.
```
drichmond: I'd say "with respect to the tile group origin"

drichmond:
This id is just to check if it's the origin tile?

libraries/bsg_manycore_cuda.cpp
bornaehsani: Oh no the origin tile has nothing to do with the tile_group_id, which goes to show how badly I've explained this in the comment.
The grid is actually just a 2D array of tile groups that represent an application, and the __bsg_tile_group_id_x/y define the index of the tile group, they are just there to identify the tile group and distinguish it from other tile groups within that grid. 
The flat __bsg_tile_group_id is just a flattened representation of the __bsg_tile_group_id_x/y, or like a 1D index of that tile group inside the grid. Maybe I should write it like this in the comment, what do you think?

drichmond: I just find 'father grid' confusing. Relative to the grid origin?

bornaehsani: OK, that sounds good, how does it look now?

libraries/bsg_manycore_cuda.cpp
```diff
@@ -2053,9 +2061,23 @@ static int hb_mc_device_tiles_set_config_symbols (hb_mc_device_t *device,
...
+		// inside the grid to distinguish it from other tile groups 
+		// __bsg_tile_group_id_x/y symbols define the X/Y coordinates of
+		// a tile's tile group relative to grid origin.
+		// The flat tile group id is calculated using the grid dimensions 
+		// and the tile group X/Y coordinates relative to grid origin
+		// __bsg_tile_group_id = __bsg_tile_group_id_y * __bsg_grid_dim_x
+		//                       + __bsg_tile_group_id_x 
+		// The flat tile group id is used in bsg_print_stat messages 
+		// to distinguish between bsg_print_stat statements of different 
+		// tile groups that run on a specific tile in sequence.
```
drichmond: bsg_tile_group_id uniquely identifies each tile group in a grid. The flat tile group id is calculated using the grid dimensions and the tile group X/Y coordinates relative to grid origin as follows: 
__bsg_tile_group_id = __bsg_tile_group_id_y * __bsg_grid_dim_x + __bsg_tile_group_id_x 

bsg_tile_group_id is used in bsg_print_stat to distinguish between unique tile group invocations that have been executed on the same tile.

drichmond: "Executed on the same tile" still trips me up, because it does not eliminate the possibility that the tile groups are executing at the same time (Not possible, I know)

libraries/bsg_manycore_cuda.cpp
```diff
@@ -2034,6 +2037,11 @@ static int hb_mc_device_tiles_set_config_symbols (hb_mc_device_t *device,
 
 		// Set tile's id __bsg_id symbol.
 		// calculate tile's id 
+		// __bsg_id symbol is the flat index of a tile 
+		// inside its tile group with respect to the origin tile
+		// __bsg_id = __bsg_y * __bsg_tiles_X + __bsg_y
+		// __bsg_tiles_X/Y symbols define the dimensions of manycore
+		// which are set to the tile group dimensions
```
drichmond: This also doesn't make sense to me. They define the dimensions of [the] manycore, and are also set to the tile group dimensions?

That seems contradictory.

drichmond:
Superseded by #466?

Issue #376, opened 2019-08-01
add sim config printout at beginning of log file / output


taylor-bsg:
Print out configuration that the simulator is using at bootup, so we know conclusively in the log exactly what was simulated.

Like this:

https://github.com/bespoke-silicon-group/bsg_manycore/blob/master/testbenches/common/v/spmd_testbench.v

drichmond:
@tommydcjung When you get a chance, can you look into if we can add this in cosim_wrapper.sv?

drichmond:
Addressed in #437

PR #437, opened 2019-10-21
Display configuration at cosimulation launch

drichmond:
Addresses #376


drichmond:
@mrutt92 This PR should be approved before #438

PR #438, opened 2019-10-21
Add support for e_mem_cfg_f1_dram

drichmond:
This enum value switches between the AXI memory model
(e_mem_cfg_default), the infinite vcache memory (e_mem_cfg_infinite)
and the DRAM model provided by F1 (e_mem_cfg_f1_dram). The latter has
the most accurate timing model, but is the slowest to simulate.

This removes the AXI_MEMORY_MODEL Makefile argument

Also removed, AXI_PROT_CHECK (stale)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,27 +65,17 @@ regression:
 
 # DEBUG=1 Opens the Waveform GUI during simulation
 
-# AXI_PROT_CHECK=1 enables an AXI Protocol Checker on the PCI
-# Interface of the AWS Shell
-
-# AXI_MEMORY_MODEL=0 directs simulation to use the slower, but more
-# accurate, DDR Model.
-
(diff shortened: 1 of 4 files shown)


mrutt92:
To be clear, you've moved the AXI_MEMORY_MODEL switch from a command line argument to a parameter? Or am I missing something?

drichmond:
Yes, I have. The reason is that switching the value of AXI_MEMORY_MODEL now requires a verilog recompilation. This changed when I last refactored the makefiles. By moving it to Makefile.machine.include, this makes re-compilation happen automatically.


It also makes it easier for you and @leonardxiang to add e_mem_cfg_sim_hbm (or whatever you want to call it)

drichmond:
@mrutt92 I added a comment describing the different enums possible

PR #434, opened 2019-10-14
Accelerate cosimulation IO

drichmond:
I've rebased the "slow clock" acceleration fix that feeds up IO speeds in cosimulation. @tommydcjung and I had this working back in August and it went stale because we never merged it.

At the moment I rebased it but it doesn't work. For example, the regression test library/test_vcache_simplified doesn't work.

PR #440, opened 2019-10-23
Fix link signal assignment at top level

leonardxiang:


drichmond:
Huh. When did this get introduced? Where does it pop up?

leonardxiang:
> Huh. When did this get introduced? Where does it pop up?

This may be a typo from previous PR #434, which passed regression in simulation.
But the mistake might pop up when you actually synthesize the design...

drichmond:
Great catch!

PR #441, opened 2019-10-24
Removed all instances of BSG_STRINGIFY, and moved to using argparsers in all tests.

drichmond:
This was... more extensive change than I expected.

* Each regression suite now defines how it runs both test_loader, and
  non test_loader tests.

* All tests use the argument parser

* BSG_MANYCORE_DIR is no longer defined (along with a few other stale
  variables)

* Removed tabs in all files I touched.

Currently passes in cosimulation. Will test in regression before merging.


drichmond:
The motivation for removing BSG_STRINGIFY is that it's a macro defined during C/C++ compilation. It's not flexible, and it's ugly. It's better to use the argument parsers.

There's only one remaining instance of it - but I call it out as deprecated.

drichmond:
Passes regression on F1

bornaehsani:
Looks great! Cosim regression passed.

PR #446, opened 2019-10-29
Fix test_softmax, add test_log_softmax, fix Float generation

save-buffer:


save-buffer:
Passes on both cosim and F1

regression/cuda/test_log_softmax.c
```diff
@@ -0,0 +1,175 @@
+#include "test_log_softmax.h"
```
bornaehsani: Could you copy the bsg_f1 license at the top of this file? You can find it at the top of all other files in bsg_f1.

regression/cuda/test_log_softmax.h
```diff
@@ -0,0 +1,14 @@
+#ifndef TEST_LOG_SOFTMAX_H
```
bornaehsani: Could you copy the bsg_f1 license at the top of this file? You can find it at the top of all other files in bsg_f1.

bornaehsani:
Other than the license the code looks great!

regression/cuda/test_log_softmax.c
save-buffer: Done!

regression/cuda/test_log_softmax.h
save-buffer: Done!

PR #453, opened 2019-11-07
Added missing synopsys translate declarations

drichmond:
The missing declarations prevent FPGA compilation


mrutt92:
oh whoops. thanks for catching that.

drichmond:
No worries - only took about 10 seconds to find and fix.

PR #454, opened 2019-11-08
Removed Makefile.machines.include from tests.mk

drichmond:
No makefile that includes tests.mk needs the variables defined in
Makefile.machines.include


drichmond:

mrutt92:
Will it pass on F1? Why are these includes unnecessary?

PR #458, opened 2019-11-12
Removed create_bladerunner_rom.py, replaced with make/shell commands

drichmond:
This change is to reduce confusion when adding new variables to the Manycore/F1 ROM.

Some people believed that the parameters lived in the create_bladerunner_rom.py (they did not), so I reduced it to a single-source of truth to reduce confusion

hardware.mk now generates the ASCII *.rom file that is used by BaseJump.

I also added documentation


drichmond:
The motivation for this commit is that I'm slowly trying to simplify the hardware build process so that we can support multiple machines. This was a sticking point when editing parameters in the past, so I'm removing it.

PR #459, opened 2019-11-13
Added preliminary support for multi-machine F1 cosimulation (WIP)

drichmond:
This is a WIP PR at the moment. I'd like to get feedback.

We're quickly approaching a point where we'll have multiple top-level files for F1 (or even FPGA) infrastructure. This is necessary because we'll be building 16x8 and 4x4 manycore designs on F1. Some of those will have 4 DDRs (A, B, C, D) and some will only have 1 DDR (C), which means different top-level files (@leonardxiang can attest to this), and different ROMs (see the PR).

What I've done here, is work with the Makefile.machine.include file to define multiple types of machines. However, instead of having multiple Makefile.machine.include files (which would vastly complicate our F1 infrastructure) there's only one Makefile.machine.include file. This file requires MACHINE_TYPE to be set, which then includes the specific machine configuration from an underlying makefile.

I like this approach because it appears similar to our previous approach. It's not a radical departure from the previous commit. It's also extensible. With a little effort, I can build multiple AGFIs for Amazon with different sizes/configurations. With each machine also defining it's own top-level file, we can support different top-level targets (E.g. the VCU 128 board).

I don't like this approach because it requires MACHINE_TYPE to be set to ensure that the correct binary is compiled/loaded. If we have multiple F1 images, people would need to switch between the different MACHINE_TYPE variables. 

Maybe this is a good motivation for only compiling one (LARGE) F1 design, and running cosimulation regression on the 4x4 (as we already do).

Changelist: 

* Moved Makefile.machine.include into hardware directory
* Created 4x4_DDRC, and 16x8_DDRC directories
* Added machine.mk fragments for defining machine parameters (used by Makefile.machine.include)
* Added Makefile variable MACHINE_TYPE that must be defined at compilation time
* Added HARDWARE_MACHINE_FILE variable that contains file-path of machine.mk

This PR only works for the library test programs (at the moment)


hardware/Makefile.machine.include
```diff
@@ -0,0 +1,47 @@
...
+$(error $(shell echo -e "$(RED)BSG MAKE ERROR: MACHINE_TYPE is not defined$(NC)"))
+endif
+
+# Confirm that MACHINE_TYPE is valid
+ifeq ($(filter $(MACHINE_TYPE),$(KNOWN_MACHINE_TYPES)),)
+$(info $(shell echo -e "$(RED)BSG MAKE ERROR: Unknown MACHINE_TYPE $(MACHINE_TYPE)"))
+$(error $(shell echo -e "$(RED)BSG MAKE ERROR: Valid MACHINE_TYPE values are $(KNOWN_MACHINE_TYPES)"))
+endif
+
+HARDWARE_MACHINE_PATH = $(HARDWARE_PATH)/$(MACHINE_TYPE)
```
drichmond: Using MACHINE_TYPE we include the appropriate machine.mk fragment from the machine directory.

hardware/hardware.mk
```diff
@@ -123,13 +131,13 @@ endef
...
 # line is a separate entry. This target generates the verilog for the rom.
-$(HARDWARE_PATH)/%.v: $(HARDWARE_PATH)/%.rom
+$(HARDWARE_MACHINE_PATH)/%.v: $(HARDWARE_MACHINE_PATH)/%.rom
 	python $(BASEJUMP_STL_DIR)/bsg_mem/bsg_ascii_to_rom.py $< \
                bsg_bladerunner_configuration > $@
 
 # This target generates the ASCII file for the ROM. To add entries to
 # the ROM, add more commands below.
-$(HARDWARE_PATH)/bsg_bladerunner_configuration.rom: $(CL_DIR)/Makefile.machine.include
+%/bsg_bladerunner_configuration.rom: %/machine.mk $(HARDWARE_PATH)/Makefile.machine.include
```
drichmond: However, these changes DO mean that we need to check for changes in machine.mk AND Makefile.machine.include

hardware/hardware.mk
```diff
@@ -151,7 +159,7 @@ $(HARDWARE_PATH)/bsg_bladerunner_configuration.rom: $(CL_DIR)/Makefile.machine.i
 # Each manycore design on has a set of parameters that define
 # it. Instead of passing these parameters as command-line defines
 # (which is tool-specific) we generate a header file.
-$(HARDWARE_PATH)/f1_parameters.vh: $(CL_DIR)/Makefile.machine.include
+%/f1_parameters.vh: %/machine.mk $(HARDWARE_PATH)/Makefile.machine.include
 	@echo "\`ifndef F1_DEFINES" > $@
```
drichmond: (Ditto above)

hardware/hardware.mk
```diff
@@ -167,7 +175,7 @@ $(HARDWARE_PATH)/f1_parameters.vh: $(CL_DIR)/Makefile.machine.include
 	@echo "\`endif" >> $@
 
 # This package defines the number of lines in the ROM
-$(HARDWARE_PATH)/bsg_bladerunner_rom_pkg.vh: $(HARDWARE_PATH)/bsg_bladerunner_configuration.rom
+%/bsg_bladerunner_rom_pkg.vh: %/machine.mk %/bsg_bladerunner_configuration.rom
```
drichmond: (Ditto above)

testbenches/simlibs.mk
```diff
@@ -139,7 +145,7 @@ VDEFINES   += COSIM
 VDEFINES   += DISABLE_VJTAG_DEBUG
 VDEFINES   += ENABLE_PROTOCOL_CHK
 
-include $(CL_DIR)/Makefile.machine.include
+include $(BSG_MACHINE_PATH)/Makefile.machine.include
```
drichmond: This is how I should have done it in the first place.

testbenches/simlibs.mk
```diff
@@ -177,9 +183,9 @@ VLOGAN_VFLAGS   += -ntb_opts tb_timescale=1ps/1ps -timescale=1ps/1ps \
 # with `make squeakyclean` or an equivalent
 $(TESTBENCH_PATH)/synopsys_sim.setup: $(TESTBENCH_PATH)/gen_simlibs.tcl
 	cd $(TESTBENCH_PATH) && vivado -mode batch -source $<
-
-$(WORKDIR)/AN.DB: $(TESTBENCH_PATH)/synopsys_sim.setup $(CL_DIR)/Makefile.machine.include $(VHEADERS) $(VSOURCES)
 	echo "$(PROJECT) : $(WORKDIR)/64" >> $(TESTBENCH_PATH)/synopsys_sim.setup
+
+$(WORKDIR)/AN.DB: $(TESTBENCH_PATH)/synopsys_sim.setup $(BSG_MACHINE_PATH)/Makefile.machine.include $(HARDWARE_MACHINE_FILE) $(VHEADERS) $(VSOURCES)
```
drichmond: HARDWARE_MACHINE_FILE is defined by Makefile.machine.include and is the file-path of machine.mk for a target machine

hardware/Makefile.machine.include
drichmond: The downside of using MACHINE_TYPE is that it needs to be defined in a file somewhere. I haven't quite figured out where the best place to define it is.

taylor-bsg:
Hi Dustin,

Maybe pull in Bandhav to take a look. I believe we had a similar system for
testing the ASIC?

drichmond:
We do! And he's on this PR.

This is slightly more complex than the manycore design because we run more than a single simv, or simv_debug program.

vb000:
@drichmond

To toss an idea, we have a different approach for a similar problem where we had to simulate multiple process corners for the tapeout. For each run (meaning for each machine config), we clone a separate run directory (separate bsg_manycore) and point the machine path to simv path to simv generated in bsg_design for each run.

Approach above needs zero flow changes and can be applied to bsg_f1 as well I think. Roughly, we can have a makefile in bsg_bladerunner that clones multiple copies of bsg_f1 each pointing to a simv of our choice. Few things about this:

- There'd be space overhead due to multiple clones. Not a serious issue if we do a shallow clone.
- Very straight forward to do parallel runs.
- Very easy to debug as each run has completely separate environment.
- Needs zero flow changes to existing infrastructure.

Here's how it was done for tapeout:
https://github.com/bespoke-silicon-group/bsg_manycore/blob/master/software/regress/Makefile

What do you think?

vb000:
The flow to compile machines with different configurations (which is main objective of this pr?); lgtm.

@leonardxiang Any specific issues preventing us from parameterizing the toplevel for different configs?

hardware/Makefile.machine.include
vb000: Define a default machine type here and leave it for user to override it on command-line?

drichmond:
> @drichmond
> 
> To toss an idea, we have a different approach for a similar problem where we had to simulate multiple process corners for the tapeout. For each run (meaning for each machine config), we clone a separate run directory (separate bsg_manycore) and point the machine path to simv path to simv generated in bsg_design for each run.
> 
> Approach above needs zero flow changes and can be applied to bsg_f1 as well I think. Roughly, we can have a makefile in bsg_bladerunner that clones multiple copies of bsg_f1 each pointing to a simv of our choice. Few things about this:
> 
> * There'd be space overhead due to multiple clones. Not a serious issue if we do a shallow clone.
> * Very straight forward to do parallel runs.
> * Very easy to debug as each run has completely separate environment.
> * Needs zero flow changes to existing cosim infrastructure.
> 
> Here's how it was done for tapeout:
> https://github.com/bespoke-silicon-group/bsg_manycore/blob/master/software/regress/Makefile
> 
> What do you think?

@vb000 In these different process corners, you had the same top-level file right?

I like the ability to parallel runs (especially in separate environments)

dpetrisko:
We do a pretty similar thing in BlackParrot except without the code clone. Separate build directories and separate run directories which symlink to the respective simv

hardware/Makefile.machine.include
drichmond: Yes, this would be a solution. 

The problem is that MACHINE_TYPE can differ between F1 and cosimulation. Is it better to define MACHINE_TYPE for each? Or is it better to define MACHINE_TYPE in one place?

That's the essential conflict of my question

drichmond:
> We do a pretty similar thing in BlackParrot except without the code clone. Separate build directories and separate run directories which symlink to the respective simv

This is also a feasible solution, but I fear a 'state explosion' problem. I posed this to @mrutt92 too. 

We have `regression` and `testbench` directories. Each directory has `specint` `cuda`, `spmd`, `library`, and `python` directories with tests. If we add M machines, then each of those directories will have M new directories for each machine - so we're adding  ~ M * 5 * 2 new directories to our tree to support different machine types.

It would certainly do the job, and it would allow us to run all our regression tests in parallel.

drichmond:
I suppose what I'm asking is - do developers, the people who might use this repository, prefer directory hierarchies, configuration files, or magic?

We've had a fair amount of debate recently about how to make this whole environment easier to use.

vb000:
> @vb000 In these different process corners, you had the same top-level file right?
> 
> I like the ability to parallel runs (especially in separate environments)

Yes, same toplevel. But that's only because we taped out one design. Need not be same toplevel, as each simv could be compiled from different toplevels.

drichmond:
Closing. WIP, may implement again later

PR #460, opened 2019-11-13
Add 3 more ddr controllers in CL

leonardxiang:
1) Added 2 additional memory configurations:
- e_vcache_blocking_axi4_f1_ddrx4_model
- e_vcache_blocking_axi4_f1_ddrx4

2) Update **bsg_axi_bus_pkg.vh** to support AXI interface transpose.

3) Add some wrappers to ease the wiring:
 - axi4_clock_converter
- axi4_transpose_h2v
- cl_sh_ddr_wrapper

4) Modify the **test_vcache_stride** test to check cache_axi_to_ddr functionality.

Depends on
bsg_manycore/feature_ddrx4_on_f1 **6b169d7**
basejump_stl/master **06d2fef**

### DIFF
--- a/hardware/bsg_axi_bus_pkg.vh
+++ b/hardware/bsg_axi_bus_pkg.vh
@@ -1,19 +1,19 @@
 // Copyright (c) 2019, University of Washington All rights reserved.
-// 
+//
 // Redistribution and use in source and binary forms, with or without modification,
 // are permitted provided that the following conditions are met:
-// 
+//
 // Redistributions of source code must retain the above copyright notice, this list
 // of conditions and the following disclaimer.
-// 
+//
 // Redistributions in binary form must reproduce the above copyright notice, this
 // list of conditions and the following disclaimer in the documentation and/or
 // other materials provided with the distribution.
-// 
+//
 // Neither the name of the copyright holder nor the names of its contributors may
 // be used to endorse or promote products derived from this software without
 // specific prior written permission.
-// 
+//
 // THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
 // ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
 // WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
@@ -28,6 +28,10 @@
 `ifndef BSG_AXI_BUS_PKG_VH
 `define BSG_AXI_BUS_PKG_VH
 
+
+// -----------------------------------------------------------
+// AXI-Lite (simple version)
+// -----------------------------------------------------------
 `define declare_bsg_axil_bus_s(slot_num_lp, mosi_struct_name, miso_struct_name) \
 typedef struct packed { \
   logic [slot_num_lp*32-1:0] awaddr ; \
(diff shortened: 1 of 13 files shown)


hardware/bsg_axi_bus_pkg.vh
```diff
@@ -57,85 +61,155 @@ typedef struct packed { \
...
+`define bsg_axil_miso_bus_width(slot_num_mp) \
+( slot_num_mp * (5 + 2*2 + 32) \
 )
 
 
-`define declare_bsg_axi_bus_s(slot_num_lp, id_width_p, addr_width_p, data_width_p, mosi_struct_name, miso_struct_name) \
+// -----------------------------------------------------------
+// AXI4 (simple version)
+// -----------------------------------------------------------
+`define declare_bsg_axi_bus_s(slot_num_mp, id_width_mp, addr_width_mp, data_width_mp, mosi_struct_name, miso_struct_name) \
```
tommydcjung: regarding this. I think it might be better to have slot_num_mp outside. If you want to make an array of axi ports, I think it's just better to do something like.
```
mosi_struct_name [slot_num_p-1:0] mosi_ports;

```

Also why do we need struct_name as macro parameter?

hardware/bsg_bladerunner_mem_cfg_pkg.v
```diff
@@ -41,21 +41,22 @@ package bsg_bladerunner_mem_cfg_pkg;
...
-    // model with low (1-2 cycle) latency in simulation. 
-    //                                      
+    // model with low (1-2 cycle) latency in simulation.
+    //
     // Either value produces the correct DDR4 Module in F1
     // implemenation
     , e_vcache_blocking_axi4_f1_dram
     , e_vcache_blocking_axi4_f1_model
-
+    , e_vcache_blocking_axi4_f1_ddrx4
```
tommydcjung: A description here would be nice.

hardware/bsg_axi_bus_pkg.vh
leonardxiang: That is a good point, and much simpler!

However, we do need some concatenated interface defined above, for the Xilinx IPs, such as crossbar...

For the second question, we need the macro name to differentiate AXI interfaces with different parameters, like the id_width_mp. It is possible that we use these AXI4 interfaces in one module.

tommydcjung: Are you saying that the ports on the xilinx IPs are defined something like,
so that you would need to transpose the array if defined the other way?

```
module xilinx_ip #(parameter slot_num_p="inv", parameter id_width_p="inv")
(
  input [slot_num_mp-1:0][id_width_mp-1:0] awid
);

endmodule
```
well in that case, this might be better.

leonardxiang: Yes, exactly.

hardware/bsg_bladerunner_mem_cfg_pkg.v
drichmond: I think @leonardxiang is going to get rid of these. I'd prefer different machine types (in directories) over different memory systems in the enum

drichmond: (But yes, comments are always useful. Tommy's feedback is spot on)

drichmond:
Hi @leonardxiang Is this PR still viable?

leonardxiang:
Hi, Dustin, I think it is obsolete.

drichmond:
@proftaylor, this is what we would need for multiple idealized memory channels.

proftaylor:
Thanks — we should discuss more with the folks who have their hands in the
code. I think what Cornell is asking for is a fake 1-cycle memory attached
to each v-cache. An interesting question is whether we could do this with
our dramsim3 infrastructure?

drichmond:
That is essentially what is being provided here, but this is limited to 3 DDR controllers.

But yes, if we can do it with dramsim3 I agree that would be a better solution.

drichmond:
Closing, superseded by HBM

Issue #456, opened 2019-11-09
Rename application_init to tile_group_enqueue


drichmond:
To me, application init suggests initialization- parsing, loading, setup. However, it looks like application init actually enqueues tilegroups onto a host queue. 

I like to suggest renaming it for clarity, since it really confused me when I tried to execute a tile group repeatedly.

taylor-bsg:
agree

bornaehsani:
A very good point, addressed in PR #461 . Thanks!

PR #461, opened 2019-11-13
Renamed hb_mc_application_init to hb_mc_kernel_enqueue 

bornaehsani:
In both library code and regression tests to avoid confusion. The function does not initialize any binaries, only enqueues the tile groups in the host to be scheduled for execution once hb_mc_tile_groups_execute is called.
Technically, the function represents a so-called application, comprised of a 2D grid of tile groups, not a singular tile group, so I left the "application" part in the name as is, and did not change it to tile_group_enqueue. With that said, I welcome any suggestions on this and other function names.
Regression tests passed COSIM.
Related to issue #456 .

### DIFF
--- a/libraries/bsg_manycore_cuda.cpp
+++ b/libraries/bsg_manycore_cuda.cpp
@@ -288,7 +288,7 @@ static int hb_mc_device_mesh_exit (hb_mc_mesh_t *mesh) {
 
 
 /**
- * Enqueues and schedules an application to be run on device
+ * Enqueues and schedules a kernel to be run on device
  * Takes the grid size, tile group dimensions, kernel name, argc, argv* and the
  * finish signal address, calls hb_mc_tile_group_enqueue to initialize all tile groups for grid.
  * @param[in]  device        Pointer to device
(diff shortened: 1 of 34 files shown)


drichmond:
Great!

I think "application" is confusing, but I don't have any great suggestions. enqueue_execution? You're executing a kernel right?

bornaehsani:
agreed, application is confusing. Yes we're running a kernel, in which case, how about kernel_enqueue? It makes perfect sense actually since this function takes in a kernel name, count and list of arguments. So in a sense it is a kernel with it's own arguments and specific grid / tile group dimensions.

drichmond:
👍

drichmond:
Good to go when cosim passes

bornaehsani:
Cosim passed, good to go!

bornaehsani:
@save-buffer FYI, since you're actively developing regression tests, the hb_mc_application_init function is renamed to hb_mc_kernel_enqueue.

save-buffer:
> @save-buffer FYI, since you're actively developing regression tests, the hb_mc_application_init function is renamed to hb_mc_kernel_enqueue.

Ah thanks for the heads up!

mrutt92:
@eafurst we should take note; this affects GraphIt.

PR #464, opened 2019-11-26
Switched to non-blocking cache as default cache

drichmond:


mrutt92:
are we satisfied with the testing for non-blocking? all the CUDA tests pass with non-blocking?

drichmond:
I ran regression, it passes. I can build an F1 image and try there too.

Dustin Richmond (He/Him)
University of California, PhD (2018)
www.dustinrichmond.com

Issue #451, opened 2019-11-06
Add cache to memory read and write test


leonardxiang:
We now have some library tests case for the cache to lv3 memory:
- test_dram_read_write
- test_manycore_vcache_sequence
- test_vcache_sequence
- test_vcache_stride
As far as I know, these tests read the latest data been written to the cache. 
I guess we need more tests to ensure that the dram has been write and read correctly by each cache column?

drichmond:
Some tests evict data from vcache, but no, in general we do not have tests in library that fully test DRAM

drichmond:
Do we have a SPMD test that does this (@tommydcjung)?

drichmond:
@mrutt92 Does #463 address this, or do we need more tests?

@tommydcjung Is there a SPMD test that we can run to test the cache?

mrutt92:
The most significant thing #463 does is it fixes test_vcache_flush, which wasn't actually doing what it was supposed to because cache parameters were hardcoded as constants rather than being read out of the ROM. As a result, test_vcache_flush was not actually reading from DRAM because it wasn't causing any evictions to occur. Now it actually does write a word of data to DRAM and read it back. AFAICT test_vcache_flush is the only test guaranteed to exercise DRAM in anyway.

The other fixes in #463 are small. test_vcache_sequence was not calculating addresses correctly. Other than that I changed the formatting for how these tests print memory and data to facilitate wave form debugging.

mrutt92:
test_vcache_sequence and test_vcache_stride both write some number of words to column caches and right now they both only exercise one cache. The difference between the two tests is that test_vcache_sequence writes to a contiguous region where as test_vcache_stride spaces the words out by a constant stride distance. It seems to me that neither test guarantees that valid data will be written to DRAM and then read back.

drichmond:
I thought one of the tests evicted data, maybe I was mistaken.

tommydcjung:
I don't think there is any spmd test to do that.

drichmond:
I believe this was addressed in #468.

PR #468, opened 2019-12-05
Made test_vcache_flush test lowest and highest word

save-buffer:
Addresses part of #451

### DIFF
--- a/libraries/bsg_manycore.cpp
+++ b/libraries/bsg_manycore.cpp
@@ -48,6 +48,7 @@
 #include <cstdio>
 #include <climits>
 #include <cstdbool>
+#include <cassert>
 #else
 #include <inttypes>
 #include <stdint.h>
@@ -56,6 +57,7 @@
 #include <stdio.h>
 #include <limits.h>
 #include <stdbool.h>
+#include <assert.h>
 #endif
 
 #include <stack>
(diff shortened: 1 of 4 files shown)


save-buffer:
Not long. Timing it gives this:
`make test_vcache_flush.log  3.96s user 0.16s system 100% cpu 4.099 total`

drichmond:
Can you raise the random address limit, and make it reproducible using a seed?

save-buffer:
Limit is now 128, seed is DEADBEEF. It takes ~64 seconds to run now in cosim.

tommydcjung:
can we also have tests to byte/half size access?

drichmond:
Thanks @tommydcjung, this is a good suggestion

tommydcjung:
If we could include tests to byte/half access in tile DMEM, that would also be helpful.

drichmond:
Also a good idea, but that would be the subject of another PR (one feature per PR policy)

drichmond:
@tommydcjung Can you file an issue for that? We'll do it

save-buffer:
Ok so I should do byte/half word testing from vcache in this PR and then from DMEM in another PR?

drichmond:
yep! Separate PR and separate test

drichmond:
@mrutt92 When you have a moment, can you add some feedback as our resident C++ guru?

regression/library/test_vcache_flush.cpp
```diff
@@ -169,15 +190,26 @@ int test_vcache_flush() {
 
         addr_bitwidth = hb_mc_config_get_network_bitwidth_addr(config);
         addrs[0] = 0;
-        addrs[1] = 1 << (addr_bitwidth-2); // Convert to word-level addr
+        addrs[NUM_TESTS - 1] = (1 << (addr_bitwidth-1)) - 1; // Word-level addr of highest word
```
mrutt92: Should this be `(1 << (addr_bitwdith)) - 1` ?

libraries/bsg_manycore.cpp
```diff
@@ -1277,9 +1277,11 @@ static int hb_mc_manycore_send_read_rqst(hb_mc_manycore_t *mc,
                 hb_mc_request_packet_set_mask(&rqst.request, HB_MC_PACKET_REQUEST_MASK_WORD);
```
mrutt92: @save-buffer I think you're already aware of this but the masks need to be set according to which bytes are being read. The NPA contains a byte address that should be used for byte selection.

regression/library/test_vcache_flush.cpp
```diff
@@ -169,15 +190,26 @@ int test_vcache_flush() {
...
+        uint32_t data32 = rand();
+        uint16_t data16 = rand() & 0xFFFF;
+        uint8_t   data8 = rand() & 0xFF;
 
-        for (dram = 0; dram < ndrams; ++dram){
+        for (dram = 0; dram < ndrams; ++dram) {
                 dram_coord_x = dram_xs[dram];
-                for(i = 0; i < NUM_TESTS; ++i){
+                for(i = 0; i < NUM_TESTS; ++i) {
                         addr = addrs[i];
```
mrutt92: Suggestion; generate the random data here so that it is not reused between tests.

PR #469, opened 2019-12-06
Adds snooping the bus to axi_to_mcl for print_stat packets 

bornaehsani:
Credit to @tommydcjung 
- Added snooping from the loader_link_sif_li/lo to intercept packets intended for bsg_print_stat and use it as input for vanilla core profiler.
- This fixes the clock domain crossing issue with the print_stat_v_i signal that is on main_clk_a0, vs. the vanilla_core_profiler which is on core_clock.
- Fixes the missing print_stat signals issue # 205 in bsg_manycore.
Cosim Regression passed.

### DIFF
--- a/hardware/bsg_print_stat_snoop.v
+++ b/hardware/bsg_print_stat_snoop.v
@@ -1,0 +1,47 @@
+module bsg_print_stat_snoop
+  import bsg_manycore_pkg::*;
+  import bsg_manycore_addr_pkg::*;
+  #(parameter data_width_p="inv"
+    , parameter addr_width_p="inv"
+    , parameter x_cord_width_p="inv"
+    , parameter y_cord_width_p="inv"
+    , parameter load_id_width_p="inv"
+
+    , parameter link_sif_width_lp=
+      `bsg_manycore_link_sif_width(addr_width_p,data_width_p,x_cord_width_p,y_cord_width_p,load_id_width_p)
+    
+  )
+  (
+    // manycore side
+    input [link_sif_width_lp-1:0] loader_link_sif_in_i
+    , input [link_sif_width_lp-1:0] loader_link_sif_out_i
+
+    // snoop signals
+    , output logic print_stat_v_o
+    , output logic [data_width_p-1:0] print_stat_tag_o
+  );
+
+  `declare_bsg_manycore_link_sif_s(addr_width_p,data_width_p,x_cord_width_p,y_cord_width_p,load_id_width_p);
+
+  bsg_manycore_link_sif_s loader_link_sif_in;
+  bsg_manycore_link_sif_s loader_link_sif_out;
+
+  assign loader_link_sif_in = loader_link_sif_in_i;
+  assign loader_link_sif_out = loader_link_sif_out_i;
+
+
+  `declare_bsg_manycore_packet_s(addr_width_p,data_width_p,x_cord_width_p,y_cord_width_p,load_id_width_p);
+  bsg_manycore_packet_s fwd_pkt;
+  assign fwd_pkt = loader_link_sif_in.fwd.data;
+
+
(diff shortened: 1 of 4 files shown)


drichmond:
This is it? No corresponding PRs in BSG Manycore?

hardware/cl_manycore.sv
```diff
@@ -348,6 +349,15 @@ module cl_manycore
...
 
+  // print stat signals for vanilla_core_profiler module
+  `declare_bsg_manycore_packet_s(addr_width_p,data_width_p,x_cord_width_p,y_cord_width_p,load_id_width_p);
+  bsg_manycore_packet_s fwd_pkt;
+  logic print_stat_v;
+  logic [data_width_p-1:0] print_stat_tag;
+  assign fwd_pkt = loader_link_sif_lo.fwd.data;
+  assign print_stat_v = loader_link_sif_lo.fwd.v & (fwd_pkt.addr == (bsg_print_stat_epa_gp >> 2)) & loader_link_sif_li.fwd.ready_and_rev;
+  assign print_stat_tag = fwd_pkt.payload.data;
+
```
drichmond: I think these should be guarded with synopsys translate on/off.

But it does beg the larger question - why not put this in the profiler itself?

tommydcjung: Because the profiler binds to vanilla_core. Also there are profilers that binds other other modules that also would use print stat signals

tommydcjung:
> This is it? No corresponding PRs in BSG Manycore?

it remains same for spmd_testbench, because io_complex is nonsynth there, and there is no CDC.

hardware/cl_manycore.sv
tommydcjung: But I think there should be a cleaner way to do this

drichmond: @tommydcjung Do you have any ideas?

drichmond:
Does this need to adapt for the AMO changes?

hardware/bsg_print_stat_snoop.v
```diff
@@ -0,0 +1,47 @@
+module bsg_print_stat_snoop
```
tommydcjung: please suggest a better name

drichmond: Snoopy? https://en.wikipedia.org/wiki/Snoopy

Jokes aside, what do you feel is wrong with the name? Happy to suggest alternatives, but I don't know what the current issue is.

hardware/cl_manycore.sv
```diff
@@ -348,6 +349,32 @@ module cl_manycore
...
+  `declare_bsg_manycore_packet_s(addr_width_p,data_width_p,x_cord_width_p,y_cord_width_p,load_id_width_p);
+  bsg_manycore_packet_s fwd_pkt;
+  assign fwd_pkt = loader_link_sif_lo.fwd.data;
+  assign print_stat_v = loader_link_sif_lo.fwd.v & (fwd_pkt.addr == (bsg_print_stat_epa_gp >> 2)) & loader_link_sif_li.fwd.ready_and_rev;
+  assign print_stat_tag = fwd_pkt.payload.data;
+  */
+  logic print_stat_v;
+  logic [data_width_p-1:0] print_stat_tag;
+
+  bsg_print_stat_snoop #(
```
drichmond: I think this should be put in the `ifdef COSIM` block below. That or it should be surrounded by a synopsys translate on/off declaration

drichmond:
@bornaehsani Can you run regression on this one last time and confirm that everything works as expected?

drichmond:
(For reference - this will be the final commit in Bladerunner v3.5.1)

bornaehsani:
> @bornaehsani Can you run regression on this one last time and confirm that everything works as expected?

COSIM regression test passed, profiler tools including stall graph and stats parser work correctly. Good to go, thank you guys!

PR #473, opened 2019-12-14
Amoswap

mrutt92:
* depends on https://github.com/bespoke-silicon-group/bsg_manycore/pull/209
* depends on https://github.com/bespoke-silicon-group/basejump_stl/pull/196
* updates mcl <-> axil interface to work with refactored Manycore network packets
* fixes a "false-positive" bug in the spmd test_loader
* adds an optional return code pointer in the cuda test_loader (needed for atomic tests)
* adds a machine config that supports AMO


mrutt92:
This is not ready yet.

PR #474, opened 2019-12-19
Small fixes and additions to test loader programs

mrutt92:
* adds a return code to the CUDA test_loader that can be ignored but can be set to non-zero to indicate an application failure.
* fixes a bug in which `ret` is shadowed and results in success being returned even on failure.


mrutt92:
passes regression

mrutt92:
It’s not needed because bsg_fail() works for the spmd tests.

PR #475, opened 2019-12-19
AMO support

mrutt92:
* modifies mcl for refactor network packet
* removes load_id_width_p where no longer applicable
* removes load_id_width_p from vcache instantiation
* removes load_id_width_p from cl_manycore_pkg
* removes load_id_width_p from non-blocking vcache instantiation
* removes load_id_width_p from bsg_bladerunner_rom
* removes deprecated cordinate ports from blocking vcache
* adds machine configurations for AMO support

### DIFF
--- a/hardware/axil_to_mcl.v
+++ b/hardware/axil_to_mcl.v
@@ -38,11 +38,10 @@ module axil_to_mcl
   , addr_width_p="inv"
   , data_width_p="inv"
   , max_out_credits_p="inv"
-  , load_id_width_p="inv"
   , rcv_fifo_els_lp=256
-  , link_sif_width_lp=`bsg_manycore_link_sif_width(addr_width_p,data_width_p,x_cord_width_p,y_cord_width_p,load_id_width_p)
-  , packet_width_lp=`bsg_manycore_packet_width(addr_width_p,data_width_p,x_cord_width_p,y_cord_width_p,load_id_width_p)
-  , return_packet_width_lp=`bsg_manycore_return_packet_width(x_cord_width_p,y_cord_width_p,data_width_p,load_id_width_p)
+  , link_sif_width_lp=`bsg_manycore_link_sif_width(addr_width_p,data_width_p,x_cord_width_p,y_cord_width_p)
+  , packet_width_lp=`bsg_manycore_packet_width(addr_width_p,data_width_p,x_cord_width_p,y_cord_width_p)
+  , return_packet_width_lp=`bsg_manycore_return_packet_width(x_cord_width_p,y_cord_width_p,data_width_p)
 ) (input                                         clk_i
   , input                                         reset_i
   // axil signals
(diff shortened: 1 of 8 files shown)


mrutt92:
* depends on https://github.com/bespoke-silicon-group/basejump_stl/pull/196
* depends on https://github.com/bespoke-silicon-group/bsg_manycore/pull/209

hardware/axil_to_mcl.v
```diff
@@ -388,24 +392,51 @@ module axil_to_mcl
...
+
+    always_comb begin
+      if (endpoint_out_packet_op == e_remote_load) begin
+        endpoint_out_packet_li[i].reg_id = fifo_req_cast[i].payload.load_id;
+      end
+      else begin
+        endpoint_out_packet_li[i].reg_id = '0;
+      end
+    end
+    always_comb begin
```
tommydcjung: This might be okay for one-load at a time, but if the host can wait for multiple loads, it wouldn't work.

mrutt92: yup; agreed. thank you.

mrutt92: actually, I'm not sure why that's true?

tommydcjung: what if remote load responses return out of order?

mrutt92: Well... if it's a load the reg_id is assigned a value from the fifo_req_cast.
When it returns, 'load_id' is set to the reg_id of returned_reg_id_r_lo.

tommydcjung: who assigns the reg_id? the API? if so, it should be fine.

hardware/axil_to_mcl.v
```diff
@@ -388,24 +392,51 @@ module axil_to_mcl
...
+      if (endpoint_out_packet_op == e_remote_store) begin
+        endpoint_out_packet_li[i].payload.data = fifo_req_cast[i].payload.data;
+      end
+      else begin
+        endpoint_out_packet_li[i].payload.load_info_s.load_info.float_wb = '0;
```
tommydcjung: To be more clear with the intent that 1-bit value is being assigned, I would suggest using 1'b0 instead.

hardware/axil_to_mcl.v
mrutt92: When the load request is sent from the host, the data field of the request packet struct is used as the load_id. The host API does set that field when performing non-blocking loads.

hardware/axil_to_mcl.v
mrutt92: cool; thank you!

hardware/axil_to_mcl.v
mrutt92: So the answer to your question is yes, the API sets it.

drichmond: @tommydcjung We handle read response reordering in software

https://github.com/bespoke-silicon-group/bsg_f1/blob/8e3086213a0b6d26c3525d87b78ba0d12ac16efa/libraries/bsg_manycore.cpp#L1547

(Credit to @mrutt92 to sending me the link to this, and implementing this)

drichmond:
This currently does not pass regression with the relevant branches (I think)

Issue #477, opened 2020-01-07
test_manycore_vcache_sequence fails, throws assert 


drichmond:
`bsg_manycore.cpp:1382: int hb_mc_manycore_write(hb_mc_manycore_t*, const hb_mc_npa_t*, const void*, size_t): Assertion 'shift == 0' failed.`

This is because it is trying to write an unaligned 4-byte word. @save-buffer, @bornaehsani and I are discussing solutions.

Filing for tracking purposes

bornaehsani:
Addressed in PR #478 .

PR #478, opened 2020-01-07
Fixes the alignment check issue for Host API

bornaehsani:
Related to issue #477 .
credit to @save-buffer and @drichmond.
- Added hb_mc_epa_check_alignment function to check for correct address alignment based on size.
- Moved all alignment functions in hb_mc_manycore_write8/16/32, hb_mc_manycore_write_mem, and hb_mc_manycore_memset function down one level to hb_mc_manycore_write function
- Modified test_manycore_vcache_sequence test write into correctly aligned addresses, and added writes to misaligned addresses to make sure the HOST API catches them.
COSIM regression passed, ready to merge.

### DIFF
--- a/libraries/bsg_manycore.cpp
+++ b/libraries/bsg_manycore.cpp
@@ -31,6 +31,7 @@
 #include <bsg_manycore_printing.h>
 #include <bsg_manycore_tile.h>
 #include <bsg_manycore_responder.h>
+#include <bsg_manycore_epa.h>
 
 #ifndef COSIM
 #include <fpga_pci.h>
@@ -1375,22 +1376,25 @@ static int hb_mc_manycore_write(hb_mc_manycore_t *mc, const hb_mc_npa_t *npa, co
         if (err != HB_MC_SUCCESS)
                 return err;
 
-        int shift = hb_mc_npa_get_epa(npa) & 0x3;
+        hb_mc_epa_t epa = hb_mc_npa_get_epa(npa);
+
+        // Check for 4-byte, 2-byte or 1-byte alignment based on size
+        if (hb_mc_epa_check_alignment(&epa, sz) != HB_MC_SUCCESS)
+                return HB_MC_INVALID;
+
+        int shift = epa & 0x3;
         /* set data and size */
         switch (sz) {
         case 4:
-                assert(shift == 0);
                 hb_mc_request_packet_set_data(&rqst.request, *(const uint32_t*)vp);
                 hb_mc_request_packet_set_mask(&rqst.request, HB_MC_PACKET_REQUEST_MASK_WORD);
                 break;
         case 2:
-                assert(shift == 0 || shift == 2);
                 hb_mc_request_packet_set_data(&rqst.request, *(const uint16_t*)vp);
                 hb_mc_request_packet_set_mask(&rqst.request, static_cast<hb_mc_packet_mask_t>(
                                                       HB_MC_PACKET_REQUEST_MASK_SHORT << shift));
                 break;
         case 1:
-                assert(shift == 0 || shift == 1 || shift == 2 || shift == 3);
                 hb_mc_request_packet_set_data(&rqst.request, *(const  uint8_t*)vp);
                 hb_mc_request_packet_set_mask(&rqst.request, static_cast<hb_mc_packet_mask_t>(
                                                       HB_MC_PACKET_REQUEST_MASK_BYTE << shift));
@@ -1773,8 +1777,6 @@ int hb_mc_manycore_write8(hb_mc_manycore_t *mc, const hb_mc_npa_t *npa, uint8_t 
  */
 int hb_mc_manycore_write16(hb_mc_manycore_t *mc, const hb_mc_npa_t *npa, uint16_t v)
 {
-        if(hb_mc_npa_get_epa(npa) & 0x1) // Check for 2-byte alignment by checking lowest bit
-                return HB_MC_INVALID;
         return hb_mc_manycore_write(mc, npa, &v, 2);
 }
 
@@ -1787,8 +1789,6 @@ int hb_mc_manycore_write16(hb_mc_manycore_t *mc, const hb_mc_npa_t *npa, uint16_
  */
 int hb_mc_manycore_write32(hb_mc_manycore_t *mc, const hb_mc_npa_t *npa, uint32_t v)
 {
-        if(hb_mc_npa_get_epa(npa) & 0x3) // Check for 4-byte alignment by checking lowest bit
-                return HB_MC_INVALID;
         return hb_mc_manycore_write(mc, npa, &v, 4);
 }
 
--- a/libraries/bsg_manycore.cpp
+++ b/libraries/bsg_manycore.cpp
@@ -615,7 +615,7 @@ static int  hb_mc_manycore_mmio_read_mmio(hb_mc_manycore_t *mc, uintptr_t offset
                 manycore_pr_err(mc, "%s: Failed: 0x%" PRIxPTR " "
                                 "is not aligned to 4 byte boundary\n",
                                 __func__, offset);
-                return HB_MC_INVALID;
+                return HB_MC_UNALIGNED;
         }
 
         addr = &addr[offset];
@@ -720,7 +720,7 @@ static int hb_mc_manycore_mmio_write_mmio(hb_mc_manycore_t *mc, uintptr_t offset
                 manycore_pr_err(mc, "%s: Failed: 0x%" PRIxPTR " "
                                 "is not aligned to 4 byte boundary\n",
                                 __func__, offset);
-                return HB_MC_INVALID;
+                return HB_MC_UNALIGNED;
         }
 
         addr = &addr[offset];
@@ -1255,6 +1255,32 @@ static int hb_mc_manycore_format_load_request_packet(hb_mc_manycore_t *mc,
 // Memory API //
 ////////////////
 
+/**
+ * Checks alignment of an epa based on data size in bytes.
+ * @param[in] epa  epa address
+ * @param[in] sz   data size in bytes.
+ * @return         HB_MC_SUCCESS if npa is aligned and HB_MC_UNALIGNED if not,
+ *                 and HB_MC_INVALID otherwise.
+ */
+static inline int hb_mc_manycore_epa_check_alignment(const hb_mc_epa_t *epa, size_t sz)
+{
+        switch (sz) {
+        case 4:
+                if (*epa & 0x3)
+                        return HB_MC_UNALIGNED;
+                break;
+        case 2:
+                if (*epa & 0x1)
+                        return HB_MC_UNALIGNED;
+                break;
+        case 1:
+                break;
+        default:
+                return HB_MC_INVALID;
+        }
+        return HB_MC_SUCCESS;
+}
+
 /* send a read request and don't wait for the return packet */
 static int hb_mc_manycore_send_read_rqst(hb_mc_manycore_t *mc,
                                          const hb_mc_npa_t *npa, size_t sz,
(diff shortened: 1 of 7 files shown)


regression/library/test_manycore_vcache_sequence.c
```diff
@@ -70,22 +70,22 @@ int test_manycore_vcache_sequence() {
...
-                for (size_t i = 0; i < ARRAY_LEN; i++) {
+                for (size_t i = 0; i < ARRAY_LEN; i += 1) {
                         if ((i % 64) == 1)
                                 bsg_pr_test_info("%s: Have written and read back %4zu words\n",
                                                  __func__, i);
                         
                         hb_mc_npa_t npa = hb_mc_npa_from_x_y(dram_coord_x,
                                                              dram_coord_y,
-                                                             BASE_ADDR+i);
+                                                             BASE_ADDR+(i << 2));
```
drichmond: I think this is a pretty minor niggle, but it would be nice if this computation `(i<<2)` were done in only one place.

regression/library/test_manycore_vcache_sequence.c
```diff
@@ -112,6 +112,56 @@ int test_manycore_vcache_sequence() {
                         
                         mismatch = mismatch || !data_match;
                 }
+
```
drichmond: I think these should be in a new regression test. With these, this test is verifying functionality of two things.

drichmond: They're functionally correct, and I like them. It doesn't make sense from an error-reporting point of view.

libraries/bsg_manycore.cpp
```diff
@@ -1375,22 +1376,25 @@ static int hb_mc_manycore_write(hb_mc_manycore_t *mc, const hb_mc_npa_t *npa, co
...
                 return err;
 
-        int shift = hb_mc_npa_get_epa(npa) & 0x3;
+        hb_mc_epa_t epa = hb_mc_npa_get_epa(npa);
+
+        // Check for 4-byte, 2-byte or 1-byte alignment based on size
+        if (hb_mc_epa_check_alignment(&epa, sz) != HB_MC_SUCCESS)
+                return HB_MC_INVALID;
+
+        int shift = epa & 0x3;
```
drichmond: I think this variable declaration can be removed, no?

libraries/bsg_manycore_epa.h
```diff
@@ -51,6 +54,33 @@ extern "C" {
...
 
+        /**
+         * Checks alignment of an epa based on data size in bytes.
+         * @param[in] epa  epa address
+         * @param[in] sz   data size in bytes.
+         * @return         HB_MC_SUCCESS if npa is aligned and HB_MC_INVALID otherwise.
+         */
+        static inline int hb_mc_epa_check_alignment(const hb_mc_epa_t *epa, size_t sz)
+        {
+                switch (sz) {
```
save-buffer: A shorter way to implement this could be to check that `sz` is 1, 2, or 4, and then check that `epa & (sz - 1) == 0`, though the way it is now may be more readable :)

libraries/bsg_manycore.cpp
```diff
@@ -1375,22 +1376,25 @@ static int hb_mc_manycore_write(hb_mc_manycore_t *mc, const hb_mc_npa_t *npa, co
...
         /* set data and size */
         switch (sz) {
         case 4:
```
drichmond: What if we define HB_MC_SIZE_32/16/8 (or similar). Naming doesn't matter, but removing the last magic numbers here to make it more obvious would be nice

libraries/bsg_manycore.cpp
```diff
@@ -1773,8 +1777,6 @@ int hb_mc_manycore_write8(hb_mc_manycore_t *mc, const hb_mc_npa_t *npa, uint8_t
  */
 int hb_mc_manycore_write16(hb_mc_manycore_t *mc, const hb_mc_npa_t *npa, uint16_t v)
 {
-        if(hb_mc_npa_get_epa(npa) & 0x1) // Check for 2-byte alignment by checking lowest bit
-                return HB_MC_INVALID;
         return hb_mc_manycore_write(mc, npa, &v, 2);
```
drichmond: The same macro values would be used here too

libraries/bsg_manycore.cpp
```diff
@@ -1787,8 +1789,6 @@ int hb_mc_manycore_write16(hb_mc_manycore_t *mc, const hb_mc_npa_t *npa, uint16_
  */
 int hb_mc_manycore_write32(hb_mc_manycore_t *mc, const hb_mc_npa_t *npa, uint32_t v)
 {
-        if(hb_mc_npa_get_epa(npa) & 0x3) // Check for 4-byte alignment by checking lowest bit
-                return HB_MC_INVALID;
         return hb_mc_manycore_write(mc, npa, &v, 4);
```
drichmond: The same macro values would be used here too

drichmond: And in the write8 case too

libraries/bsg_manycore_epa.h
```diff
@@ -51,6 +54,33 @@ extern "C" {
...
+                case 4:
```
drichmond: Macro values would be used here too

libraries/bsg_manycore_epa.h
drichmond: Yes, but then you'd need to validate sz.

libraries/bsg_manycore_epa.h
mrutt92: I'm not sure I agree that a macro value should be used here. @drichmond why do you think a macro should be used?

drichmond: Here, specifically? Or in general on the 4/2/1 values? 

My opinion is that we're using 4/2/1 repeatedly in this code. We use it in the writeN functions, we use it in the alignment checker, and we use it in the write() function. It feels gratuitous to me, hence the macro

libraries/bsg_manycore.cpp
mrutt92: Personally, I don't really see the point. There's 0% chance we would ever change the values of those constants. It's almost like having a macro like this:
```
#define ONE 1
```
I also don't think it makes it more legible.

drichmond: #define SGTM 1

regression/library/test_manycore_vcache_sequence.c
```diff
@@ -70,22 +70,22 @@ int test_manycore_vcache_sequence() {
...
+                                                             BASE_ADDR+(i << 2));
                         write_data = rand();
                         err = hb_mc_manycore_write_mem(mc, &npa, &write_data, sizeof(write_data));
                         if (err != HB_MC_SUCCESS) {
                                 bsg_pr_err("\n%s: failed to write A[%d] = 0x%08" PRIx32 ""
                                            " to DRAM coord (%d,%d) @ 0x08%" PRIx32 ": %s\n",
                                            __func__, i, write_data,
                                            dram_coord_x, dram_coord_y,
-                                           (npa.epa+i) << 2,
+                                           (npa.epa+ (i << 2)) << 2,
```
mrutt92: I'm not sure why you left shift twice here... I think the EPA field stored as a byte address?

regression/library/test_manycore_vcache_sequence.c
```diff
@@ -96,7 +96,7 @@ int test_manycore_vcache_sequence() {
                                            "from DRAM coord (%d,%d) @ 0x%08" PRIx32 ": %s\n",
                                            __func__, i,
                                            dram_coord_x, dram_coord_y,
-                                           (npa.epa+i) << 2,
+                                           (npa.epa+ (i << 2)) << 2,
```
mrutt92: See my note above.

regression/library/test_manycore_vcache_sequence.c
drichmond: Yes, you're right. i needs to be shifted (because it's a word index, not a byte address)

mrutt92: Yes, I understanding shifting i, but I don't understand shifting the entire resulting address.

libraries/bsg_manycore_epa.h
mrutt92: Are you suggesting this code become
```
case FOUR:
 // ...
case TWO:
// ...
case ONE:
// ...
```
Or am I misunderstanding you?

drichmond: I agree with you, I was just making a (bad) joke

mrutt92: OK, I wasn't sure if this conversation was also resolved.

regression/library/test_manycore_vcache_sequence.c
drichmond: I'm just writing poor responses right now...

Yes, I agree with you. I'll fix this.

regression/library/test_manycore_vcache_sequence.c
mrutt92: I also think that we should better than just testing the base address.

drichmond: Can you file an issue?

drichmond: I think that Sasha's new test does more than this. I think it tests the high and low address of each vcache

regression/library/test_manycore_vcache_sequence.c
```diff
@@ -112,6 +112,56 @@ int test_manycore_vcache_sequence() {
...
+                err = hb_mc_manycore_write_mem(mc, &npa, &write_data, sizeof(write_data));
+                if (err != HB_MC_SUCCESS) { 
+                        bsg_pr_test_info("%s: Expected an invalid address error. We're good!\n", __func__);
+                        bsg_pr_err("%s: Did not catch write into invalid mis-alined address: 0x%08" PRIx32 ".\n",
+                                   __func__, hb_mc_npa_get_epa(&npa));
+
+                }
+                else {
+                        bsg_pr_err("%s: Did not catch write into invalid mis-alined address: 0x%08" PRIx32 ".\n",
+                                   __func__, npa);
```
mrutt92: You want `hb_mc_npa_get_epa(&npa)` here.

drichmond:
OK, so I fixed the comments. 

* Added alignment checks in read methods
* Added new test for alignment that also tests reads and the read/writeN versions
* Fixed shifting issues in vcache_sequence (not a functional bug)

regression/library/test_manycore_vcache_sequence.c
mrutt92: *sigh* if only stdio were type-safe...

mrutt92: separate issue, we should file an issue to turn on -Wall -Werror in the regression tests.

drichmond: /giphy can it. 

(My magic abilities only extend so far)

drichmond: Yes, to -wall/-werror.

drichmond: I think I did at some point. I think the loader test broke

drichmond: The underlying issue here is fixed

mrutt92: Turning on -wall -werror is never the trivial task it should be...

regression/library/test_manycore_alignment.c
```diff
@@ -0,0 +1,327 @@
...
+
+        uint32_t dram_coord_x = 0;
+
+        uint32_t dram_coord_y = hb_mc_config_get_dram_y(config);
+        int mismatch = 0;
+
+        /*
+         * Loop over all DRAM banks and try to read/write to an unaligned address
+         */
+        for (dram_coord_x = 0; dram_coord_x < hb_mc_dimension_get_x(dim); dram_coord_x++) {
```
mrutt92: Even though these are all supposed to return error, some random testing would be good.

mrutt92: As in random addresses. I can see that the data is random.

regression/library/test_manycore_vcache_sequence.c
mrutt92: Different test with different consequences and different meaning.

drichmond: So, this one just does sequences. If you want sequences starting from *random* locations, the world is your oyster

mrutt92: This thread is out of date.

drichmond: Join was called

PR #480, opened 2020-01-10
Find and fix bug in test_vcache_flush and library

save-buffer:


drichmond:
Looks good, but I don't understand what the bug was. Can you describe?

save-buffer:
Essentially, the mask doesn't do anything for reads and only affects writes. When you send a read packet requesting a word, the manycore will send the entire word, regardless of what the mask is. As a result, you have to shift it down. 
Similarly, when you send a write packet, you don't send the word you want in the low-order bits, you have to shift that up to the proper position along with the mask.

drichmond:
@mrutt92 LGTY?

libraries/bsg_manycore.cpp
```diff
@@ -1363,22 +1363,22 @@ static int hb_mc_manycore_recv_read_rsp(hb_mc_manycore_t *mc,
         }
 
         /* check that the npa matches? */
-
+        
+        int shift = CHAR_BIT * (hb_mc_npa_get_epa(npa) & 0x3);
```
mrutt92: Technically CHAR_BIT is the number of bits in a byte on the host machine ;-)

drichmond: So, sizeof(char)*8?

mrutt92: Nah, it's fine as is! I'm just giving a hard time.

PR #481, opened 2020-01-12
Add 2d convolution test

save-buffer:
See https://github.com/bespoke-silicon-group/bsg_manycore/pull/210 for kernel

### DIFF
--- a/regression/cuda/test_conv2d.c
+++ b/regression/cuda/test_conv2d.c
@@ -1,0 +1,283 @@
+// Copyright (c) 2019, University of Washington All rights reserved.
+// 
+// Redistribution and use in source and binary forms, with or without modification,
+// are permitted provided that the following conditions are met:
+// 
+// Redistributions of source code must retain the above copyright notice, this list
+// of conditions and the following disclaimer.
+// 
+// Redistributions in binary form must reproduce the above copyright notice, this
+// list of conditions and the following disclaimer in the documentation and/or
+// other materials provided with the distribution.
+// 
+// Neither the name of the copyright holder nor the names of its contributors may
+// be used to endorse or promote products derived from this software without
+// specific prior written permission.
+// 
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+// ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+// WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+// DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR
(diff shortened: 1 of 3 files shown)


regression/cuda/test_conv2d.c
```diff
@@ -0,0 +1,283 @@
...
+#define MAX_S 8
+
+#define MAX_VALUE_MAGNITUDE 1e6
+
+int generate_inclusive(int min, int max)
+{
+        int range = (max - min + 1);
+        return (rand() % range) + min;
+}
+
```
drichmond: You should see how I've started generating random numbers here: https://github.com/bespoke-silicon-group/bsg_f1/blob/9cdfa2862ca547dc6d9b5fc817db1bf66ba70403/regression/cuda/test_vector_matrix_mul.cpp#L95

It's worth switching to C++ to get nice things :p

regression/cuda/test_conv2d.c
```diff
@@ -0,0 +1,283 @@
...
+int output_dim(int N, int F, int P, int S)
+{
+        return 1 + (N - F + 2 * P) / S;
+}
+
+// Takes an MxN matrix A and a HxW filter, a padding P,
+// a vertical stride Sy and a horizontal Sx, and outputs
+// the result of a 2D convolution into B. B must be of size
+// output_dim(M, H, P, Sy) x (N, W, P, Sx)
+void conv2d(const float *A,
```
drichmond: Likewise for here - you've written this for float, but we're also interested in other datatypes. Writing this in C++ with templates would make it a bit more general

regression/cuda/test_conv2d.c
```diff
@@ -0,0 +1,283 @@
...
+
+        int M = generate_inclusive(MIN_N, MAX_N);
+        int N = generate_inclusive(MIN_N, MAX_N);
+        int H = generate_inclusive(MIN_F, MAX_F < N ? MAX_F : N);
+        int W = generate_inclusive(MIN_F, MAX_F < N ? MAX_F : N);
+        int P = generate_inclusive(MIN_P, MAX_P);
+        int Sx = generate_inclusive(MIN_S, MAX_S);
+        int Sy = generate_inclusive(MIN_S, MAX_S);
+        int Bx = output_dim(M, H, P, Sy);
+        int By = output_dim(N, W, P, Sx);
```
drichmond: In the code I linked previously I generate for doubles, and then convert to float/int/int16/int8

regression/cuda/test_conv2d.c
save-buffer: Hmmm I'm not convinced C++ would be very much nicer in this particular case - since the limits are different each time I generate it, I'd have to make a new generator object every time, and then I'd have to have a random seed each time.

drichmond:
How long should this test take to run?

drichmond:
This passes, but takes way too long to run.

save-buffer:
Oh yeah it took me like 3 hours... I'll decrease the matrix size.

drichmond:
How long does it take now?

save-buffer:
`make test_conv2d.log  75.60s user 0.21s system 99% cpu 1:15.82 total`

PR #482, opened 2020-01-14
adjusting dram size params

tommydcjung:
BSG_MACHINE_DRAM_SIZE_WORDS = how big is the main memory? The max is 2GB, but this value can be set to lower (e.g. 512MB).

BSG_MACHINE_DRAM_BANK_SIZE_WORDS = how much address space does each bank (vcache) cover? This should be BSG_MACHINE_DRAM_SIZE_WORDS divided by the number of columns.


drichmond:
Does this pass in cosim for both machine files?

drichmond:
Effectively closing this, but leaving it open because the numbers are relevant for later. Ramulator support is deprecated in favor of DramSim3

Issue #476, opened 2020-01-06
write a regression test to exercise 8 and 16 bit reads/writes (to DMEM)


mrutt92:
Please write a simple regression test to exercise 8 and 16 bit reads/writes.

PR #485, opened 2020-01-16
Add test for 8- 16- and 32-bit dmem reads/writes

save-buffer:
Addresses #476


drichmond:
Looks good to me! 

Passes cosimulation regression?

save-buffer:
Yes!
image

PR #492, opened 2020-01-20
Add -Wall and -Werror to cosimulation compile flags

drichmond:
This PR adds -Wall and -Werror to compile flags.

However, I'm not entirely comfortable with the changes that were necessary (though simple)

To summarize: 
* VCS doesn't have a CXXFLAGS option, only CFLAGS
* VCS uses gcc by default as its compiler to compile c-generated source files
* VCS must use g++ not gcc avoid compile-time errors
* VCS generated C source isn't C++ compliant (typecasting from int* to unsigned int*)

Therefore, the solution is to modify the DPI interface slightly (using longint unsigned instead of unsigned int), which triggers a void* instead of an int* (when we wanted an unsigned int*)

This passes cosimulation, but I am fairly leery about this PR until @mrutt92 and I talk it through.

(Adding @dpetrisko for input on Verilog types)

### DIFF
--- a/regression/cuda/tests.mk
+++ b/regression/cuda/tests.mk
@@ -85,6 +85,6 @@ DEFINES += -D_XOPEN_SOURCE=500 -D_BSD_SOURCE
 CDEFINES   += $(DEFINES)
 CXXDEFINES += $(DEFINES)
 
-FLAGS     = -g -Wall
+FLAGS     = -g -Wall -Werror
 CFLAGS   += -std=c99 $(FLAGS) 
 CXXFLAGS += -std=c++11 $(FLAGS)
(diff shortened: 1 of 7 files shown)


drichmond:
Solves #492

dpetrisko:
DPI is explicitly C code in the standard (not c++), and there’s a table mapping the Verilog types to C types that I can dig up to verify the correct types are being used. 

Priority to me is being compatible with the Verilog spec, then Verilog tools, then C/C++ tools.   So I would say using the correct types and waiving the C warnings through casts or pragmas is preferable.

drichmond:
That's a good hierarchy. 

We don't have any control over the generated C code because it's generated by VCS. Max and I tried to find compiler flags that would turn off the two warnings that are causing issues, but we couldn't.

There are other options though. For example, I could compile the C++ file into a .o file before calling VCS (which would use gcc). This would avoid the problem all together (I think). However, that's more disruptive on our current scripts.

dpetrisko:
Great minds think alike, that’s what I was going to suggest — compiling and linking separately. I’m not familiar enough with the cosim process to weigh in how reasonable a change that is, though.

drichmond:
Closing, see new branch wall-werror-v2

PR #494, opened 2020-01-21
Part 1 of Enable -Wall/-Werrror

drichmond:
Enabling -Wall/-Werror is more complex than we initially anticipated, but it has great benefits. Since enabling -Werror causes many of the regression tests to fail during compilation, I've split this into two parts.

This PR enables -Werror to be set, which was not possible previously.

This PR unifies "compilation" (.o file generation) for F1 and Cosimulation frameworks, and reduces VCS involvement to linking ONLY. It may simplify long-term python integration (@vb000, @sampsyo)

* C/C++ files are now compiled by gcc/g++ into .o files (instead of opaquely with VCS)
* Replaced testbenches/compilation.mk to link.mk (with VCS commands)
* Split (part) of regression/compilation.mk into link.mk
* Renamed two .h files to .hpp

(Currently passes cosimulation)


drichmond:
This PR isn't quite complete: Documentation and Error-checking still required, but the idea is complete.

drichmond:
PR is complete

mrutt92:
I added to a couple makefiles. Looks good.

PR #495, opened 2020-01-22
enum type fix

tommydcjung:


hardware/axil_to_mcl.v
```diff
@@ -394,7 +394,7 @@ module axil_to_mcl
     assign fifo_req_cast[i]          = fifo_data_lo[2*i];
 
     bsg_manycore_packet_op_e endpoint_out_packet_op;
-    assign endpoint_out_packet_op = fifo_req_cast[i].op;
+    assign endpoint_out_packet_op = bsg_manycore_packet_op_e'(fifo_req_cast[i].op);
```
mrutt92: why is this necessary?

drichmond: Vivado synthesis fails with the following error

```ERROR: [Synth 8-1587] an enum variable may only be assigned to same enum typed variable or one of its values [/mnt/users/spin0/no_backup/drichmond/Research/repositories/git/bsg_bladerunner.bak/bsg_f1/build/src_post_encryption/axil_to_mcl.v:397]```

mrutt92: got it. thanks!

tommydcjung: This also causes lint warning in vcs. We should do our best to pay attention to those warnings and fix them.

PR #497, opened 2020-01-24
Make Flow modifications

drichmond:
* Removed a few unnecessary checks
* Moved rule declarations to more relevant files
* Added BSG_MACHINE_NAME to all Makefile.machine.include

@vb000 @tommydcjung - Can you give input on the new BSG_MACHINE_NAME variable? I'm using this to build machine-specific crt.o files so that someone can't run with the wrong crt.o file for a machine, among other things.

@mrutt92 for LGTM

### DIFF
--- a/Makefile.machine.include
+++ b/Makefile.machine.include
@@ -38,7 +38,9 @@ BSG_MACHINE_DRAM_SIZE_WORDS           = 536870912
 BSG_MACHINE_DRAM_BANK_SIZE_WORDS      = 134217728
 
 BSG_MACHINE_DATA_WIDTH                = 32
-BSG_MANYCORE_MACHINE_NAME             := Y$(BSG_MACHINE_GLOBAL_Y)X$(BSG_MACHINE_GLOBAL_X)_NB_DDR
+
+# BSG Machine Name is used to build non-conflicting crt.o files
+BSG_MACHINE_NAME                      = Y$(BSG_MACHINE_GLOBAL_Y)X$(BSG_MACHINE_GLOBAL_X)_NONBLK_DDR4_F1MODEL
 
 # This flag has to be always 0 by default. Conditional
 # assignment allows user to set this flag through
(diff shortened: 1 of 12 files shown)


Makefile.machine.include
```diff
@@ -39,6 +39,9 @@ BSG_MACHINE_DRAM_BANK_SIZE_WORDS      = 134217728
 
 BSG_MACHINE_DATA_WIDTH                = 32
 
+# BSG Machine Name is used to build non-conflicting crt.o files
+BSG_MACHINE_NAME                      = Y$(BSG_MACHINE_GLOBAL_Y)X$(BSG_MACHINE_GLOBAL_X)_NONBLK_DDR4_F1MODEL
```
tommydcjung: I think it should be somewhat similar format as mem cfg names.
e.g.) y12_x8_vcache_non_blocking_ddr4_f1model

tommydcjung:
why should the crt.o vary with different machine types?

drichmond:
It was my impression that it linking against the wrong crt.o sometimes causes hangs: For example, Some machines have DRAM, some use Vcache. Some that have DRAM use different memory sizes.

Makefile.machine.include
drichmond: I think this is good feedback. I'd be happy to do this too. 

I deferred to shorter names, but I'd be happy with your suggestion too.

vb000:
crt has these implemented as symbols, afaik. Will take a look why these have to be different.

drichmond:
It’s partly because of the .data location and stack pointer, right?

vb000:
crt uses `_sp` symbol defined in the link script to initialize sp. So, should be independent of that...

drichmond:
Hmmm. There’s no other arch specific part of crt.o? Is it just the link script?

Aside, can you rename link_dram2/dmem2 to be a little more descriptive?

vb000:
No -- but I would rely on regression to prove it. Yes, it's just the link script that changes. Sure, I'll rename them.

Makefile.machine.include
drichmond: I'm going to roll back the name addition. @vb000 is right, recompilation is not needed for crt.o

drichmond:
Passes cosimulation on the default Makefile.machines.include and on 4x8_non_blocking_vcache_ramulator_hbm.

Also passes regression on F1

PR #498, opened 2020-01-27
Pytorch setup

vb000:
- Flow to detect anaconda environment and adapt python flags according to that.
- A couple of pytorch test programs.


hardware/hardware.mk
```diff
@@ -128,7 +128,7 @@ endef
 # data. Each rom entry is a 32-character string with 1/0 values. Each
 # line is a separate entry. This target generates the verilog for the rom.
 $(HARDWARE_PATH)/%.v: $(HARDWARE_PATH)/%.rom
-	python $(BASEJUMP_STL_DIR)/bsg_mem/bsg_ascii_to_rom.py $< \
+	/usr/bin/python $(BASEJUMP_STL_DIR)/bsg_mem/bsg_ascii_to_rom.py $< \
                bsg_bladerunner_configuration > $@
```
drichmond: Why is it necessary to use /usr/bin? That's a bit too OS-specific for my liking

regression/python/tests.mk
```diff
@@ -46,7 +48,16 @@ DEFINES += -D_XOPEN_SOURCE=500 -D_BSD_SOURCE
...
+CONDA_PREFIX := $(shell echo $$CONDA_PREFIX)
+
+ifeq ($(CONDA_PREFIX),)
+  FLAGS     = -g -Wall $(shell python3-config --cflags) -O1 
+  LDFLAGS  += $(shell python3-config --ldflags)
+else
+  FLAGS     = -g -Wall -I$(CONDA_PREFIX)/include/python3.7m -O1 
+  LDFLAGS  += -L$(CONDA_PREFIX)/lib -lpython3.7m -Wl,-rpath,$(CONDA_PREFIX)/lib
+endif
+
```
drichmond: Is it worth it to have a separate regression directory for pytorch?

hardware/hardware.mk
vb000: basejump scripts tend to use system python (we can say python2). So, using system python felt right. Agree, but it's what the python file intends to do: https://github.com/bespoke-silicon-group/basejump_stl/blob/0e3313335586c399c839249364fc209a6b45cb74/bsg_mem/bsg_ascii_to_rom.py#L1.

regression/python/tests.mk
vb000: This part is not pytorch specific. This logic is just to distinguish between system python and anaconda's python...

hardware/hardware.mk
vb000: Oh, and the actual reason I had to change is, when I'm in python3 environment, basejump_stl's python2 scripts fail.

drichmond: A better solution might be `env python2` or `env python3`. That is path independent 

See: https://stackoverflow.com/questions/21612980/why-is-usr-bin-env-bash-superior-to-bin-bash

regression/python/tests.mk
drichmond: I think there is value in splitting the pytorch regression tests into a separate regression directory. They depend on the anaconda environment, right?

vb000: Definitely, I also have been thinking that having separate directories would make everything easy to understand. Sure, I'll add a separate pytorch regression directory.

hardware/hardware.mk
vb000: That means we would replace `python` with `python2` as the executable name? 

Another idea to treat the script as an executable and and let it handle which interpreter it invokes. Does this sound good?

drichmond: Sounds good to me, as long as the script doesn't hard code to use `/bin/env`

vb000: Actually, it does. And, I think it's okay because the script is written for that interpreter? I understand the executable, rarely, might not be there, but not on any linux. 

But since python2 is officially deprecated, we might not find python2 executable in modern python distributions. So, explicitly searching for python2 binary using `env python2` is also a great idea.

@proftaylor, a note about python2 use in basejump_stl.

vb000:
Addressed both issues...

drichmond:
Looks good to me. I'm going to wait to merge until after v3.6.0 so that I can apply any fixes (probably tomorrow)

PR #500, opened 2020-01-28
Cleaned up some make variables

drichmond:
* Used BSG_MACHINE_PATH instead of CL_DIR to unify with Manycore flow
* Replaced $(CL_DIR)/hardware with $(HARDWARE_PATH)
* Moved definition of WRAPPER_NAME


drichmond:

Issue #253, opened 2019-06-06
Add "fence" method to Host API


drichmond:
All the host method should do is wait until the credits return to the maximum value. It can have a timeout as well.

I think this should be used in _exit initially. Others can follow.

drichmond:
Depends on #250

tommydcjung:
how does the host check if the credits return to the max value? does it poll the credit value periodically?

tommydcjung:
alternative is to implement a fence packet which blocks the pipeline until the credit counter returns to max

drichmond:
Yes, it would poll the credit value periodically.

drichmond:
But I’m open to other suggestions

tommydcjung:
I think this should be all handled in hardware. software/host shouldn't have to worry about how many credits are left.

drichmond:
So a fence would be sort of like a request packet - It would go out, and the response wouldn’t be enqueued until the credits were returned

tommydcjung:
host should be able to sent load, store, fence packet. fence blocks all the subsequent packets until the credit counter is fully restored. fence is not actually sent out to the network.

tommydcjung:
after host sends the fence, host should still be able to send other load/store packets. to take advantage of the big fifo..

drichmond:
Yeah, that’s what I meant. The fence would never leave the host interface.

drichmond:
This sounds a bit like multithreading, which we don’t currently support. Is that accurate?

tommydcjung:
when the host program sends load for example, is it blocking or non-blocking? presumably, the host can send out multiple requests and get the response asynchronously

drichmond:
We batch, but it’s blocking. Multiple loads (or stores) can be in flight at the same time, but they’re all from the same method call

max-ruttenberg:
@drichmond addressed in https://github.com/bespoke-silicon-group/bsg_f1/pull/249?

drichmond:
No - there should be an explicit low-level driver method that we can call that blocks until the credits return to maximum.

drichmond:
@tommydcjung and I were debating how this would happen - for example, we could send a special "fence" packet that would cause the host interface to block sending requests until the credits are all returned, and then return the fence packet to the host. Or we could block in software. 

There are benefits and drawbacks to both.

drichmond:
@leonardxiang we should also be able to read remaining credits in your pr

[...]

PR #503, opened 2020-02-06
Remove Xilinx IP in the manycore link

leonardxiang:
Refactor the manycore link with basejump_stl:

- RTL simplification (Thanks to @tommydcjung ):
         bsg_manycore_link_to_axil
          --  bsg_mcl_axil_fifos_master (for host write and read)
          --  bsg_mcl_axil_fifos_slave (for manycore write)
          --  bsg_manycore_endpoint_to_fifos

- API changes:
        -- MMIO address mapping has been redefined for simplicity. see Doc.
        -- Manycore init: I removed the interrupt&status register(ISR), also removed the rx fifo vacancy register for the manycore response path
        -- Host write: Host updates its write credits by checking the vacancy of the tx fifo.
        -- Host read: Host tracks its credit, which is 32 (unchanged, limited by the reg_id width)
        -- Host write request fence: Host checks both the vacancy of the tx fifo and the out credits of the endpoint standard to ensure that all requests has been returned.
        -- manycore write: host first checks the rx fifo occupancy, then de-queue the fifo by axil read.

- All regressions pass on cosim and FPGA board.

- Dependency: bsg_bladerunner v3.6.0

- Datasheet:
https://docs.google.com/document/d/1-jSBELaYREEqtGOUD4_yAnl4EJRAZ0hOJfu-_1ZQllw


libraries/bsg_manycore.cpp
```diff
@@ -392,29 +346,14 @@ static int hb_mc_manycore_init_config(hb_mc_manycore_t *mc)
...
-}
 
 /*
  * These might be rewritten to read from MMIO space: hence why error codes are returned.
  */
 static int hb_mc_manycore_get_host_requests_cap(hb_mc_manycore_t *mc, unsigned *cap)
 {
-        *cap = 32;
+        // *cap = HB_MC_MMIO_MAX_READ_CREDITS;
+        *cap = 32;  // corresponding to the reg_id width
```
drichmond: Why is this hard coded?

libraries/bsg_manycore.cpp
```diff
@@ -504,9 +421,8 @@ static int hb_mc_manycore_init_fifos(hb_mc_manycore_t *mc)
 
 static void hb_mc_manycore_cleanup_fifos(hb_mc_manycore_t *mc)
 {
-        /* drain rx FIFOs */
+        /* drain rx FIFO from the manycore request */
```
drichmond: Drain the Manycore-To-Host (RX) Request FIFO

libraries/bsg_manycore.cpp
```diff
@@ -682,7 +597,7 @@ int hb_mc_manycore_get_host_credits(hb_mc_manycore_t *mc)
         uint64_t addr;
         uint32_t value;
         int err;
-        addr = hb_mc_mmio_credits_get_reg_addr(HB_MC_MMIO_CREDITS_HOST_OFFSET);
```
drichmond: Why did you remove this macro? It was useful for abstracting the address space

libraries/bsg_manycore.cpp
```diff
@@ -1247,6 +1082,7 @@ static int hb_mc_manycore_format_load_request_packet(hb_mc_manycore_t *mc,
         if ((r = hb_mc_manycore_format_request_packet(mc, pkt, npa)) != 0)
                 return r;
 
+        hb_mc_request_packet_set_data(pkt, 0x0);
```
drichmond: What is 0x0?

libraries/bsg_manycore.cpp
```diff
@@ -1306,7 +1142,8 @@ static int hb_mc_manycore_send_read_rqst(hb_mc_manycore_t *mc,
                 return err;
 
         // mark request with id
-        hb_mc_request_packet_set_data(&rqst.request, id);
+        // hb_mc_request_packet_set_data(&rqst.request, id);
```
drichmond: Can you remove this commented line, if it's no longer relevant?

libraries/bsg_manycore_request_packet.h
```diff
@@ -218,14 +219,24 @@ extern "C" {
                 packet->y_src = y;
         }
 
+        /**
+         * Set the read id in a request packet
+         * @param[in] packet a request packet
+         * @param[in] rd for for amo and int/float load
+         */
+        static inline void hb_mc_request_packet_set_reg_id(hb_mc_request_packet_t *packet, uint8_t reg_id)
```
drichmond: Why is it called "set_reg_id?"

libraries/bsg_manycore.cpp
```diff
@@ -467,33 +406,11 @@ static int hb_mc_manycore_init_fifos(hb_mc_manycore_t *mc)
...
-        rc = hb_mc_fifo_enable(mc, HB_MC_MMIO_FIFO_TO_DEVICE);
-        if (rc != HB_MC_SUCCESS)
-                return rc;
-
-        rc = hb_mc_fifo_enable(mc, HB_MC_MMIO_FIFO_TO_HOST);
-        if (rc != HB_MC_SUCCESS)
-                return rc;
-
-        /* drain rx FIFOs */
+        /* drain rx FIFO from the manycore request */
```
drichmond: Drain the Manycore-to-Host (RX) Fifo

libraries/bsg_manycore.cpp
drichmond: Can it be put in the rom?

libraries/bsg_manycore.cpp
```diff
@@ -120,53 +120,14 @@ static int hb_mc_manycore_packet_rx_internal(hb_mc_manycore_t *mc,
...
-        if (err != HB_MC_SUCCESS) {
-                manycore_pr_err(mc, "Failed to get %s vacancy\n", typestr);
-                return err;
-        }
-
-        return HB_MC_SUCCESS;
+        manycore_pr_dbg(mc, "We do not check the %s FIFO's vacancy\n", typestr);
+        vacancy = NULL;
+        return HB_MC_FAIL;
 }
```
drichmond: Why not? It seems like poor form to return HB_MC_FAIL in all cases. If you don't expect this method to be used anymore it should be removed

libraries/Makefile
```diff
@@ -57,7 +57,7 @@ install: $(LIBRARIES_PATH)/libbsg_manycore_runtime.so.1.0
 	mv $(notdir $<) /usr/lib64/
 	ln -sf /usr/lib64/$(notdir $<) /usr/lib64/libbsg_manycore_runtime.so.1
 	ln -sf /usr/lib64/$(notdir $<) /usr/lib64/libbsg_manycore_runtime.so
-	cp -t /usr/include $(HEADERS)
+	cp -t /usr/include $(LIB_HEADERS)
```
drichmond: Can you rebase on master? I believe this has changed

drichmond:
Thank you @leonardxiang! I focused on the C/C++ code, and I hope that @mrutt92 has time to do the same.

drichmond:
But as a general comment, why was flow control removed from software? How does the software know that it won't overflow the transmit fifo? Does it track credits?

leonardxiang:
> But as a general comment, why was flow control removed from software? How does the software know that it won't overflow the transmit fifo? Does it track credits?

Good question.
The flow control for host request is now being handled in hardware. If the tx fifo is full, then the AXIL write will be stalled. Actually, after discussing with @tommydcjung, I believe that the requests will always be absorbed by the network, eventually. 

In conclusion, the host only does 2 things for flow control. One is tracking the credits when doing remote load. The other is polling the Manycore-to-Host (RX) Fifo before read. 

Please refer to the Datasheet for details.

drichmond:
But if the AXI Lite write is stalled, doesn’t that stall the PCIE bus? 


Dustin Richmond (He/Him)
University of California, PhD (2018)
www.dustinrichmond.com

leonardxiang:
I think it will. 
We "believe" that the network throughput is greater than the IO, and it will not "stall" for actual program. 
So, what implementation do you prefer?

drichmond:
I don't know. Does cosimulation overflow the capacity, and it doesn't fail? 

I believe that you're 99.9% correct for PCIe. IO bandwidth will never exceed the network bandwidth on a PCIe device - but that may only be true for PCIe. Since you never know what this might be reused in, I would be strongly in favor of SOME flow control. 

For example, what happens if we were to use this with Black Parrot? Maybe that is able to issue a write on every cycle at a rate faster than the manycore can handle.

leonardxiang:
The corner case of tx fifo full is observed in the cosim.

The question is: is there such a time-out setting for the PCIe-to-AXI access.
The easiest way is to expose a tx fifo vacancy register to the user, so that it can check the available space before batching the requests, with price of losing bandwidth and simplicity.

drichmond:
Why can’t we use the credits for flow control? Does the host software read them?

dpetrisko:
Hi, BlackParrot also uses similar credit-based hardware flow control.  I believe it would just result in back pressure. (Not to say that it would not also benefit from software control for performance, but it should not cause correctness problems).

taylor-bsg:
Thanks Leonard; can you make a copy of the doc in your personal gmail
account, and then assign ownership to me? UW will delete this doc shortly
after your graduation.

taylor-bsg:
The delay on the manycore network can be arbitrarily long although never
infinite. So if PCI-E is going to get unhappy because of a stall that is
too long (e.g 5000 ns), then some kind of credit counting may be necessary.
Maybe a cached copy of the credit counter can be used, and only when it
hits zero is the PCIe register read and the cached copy updated.

drichmond:
To summarize, what I believe the current concerns are: 

- There's no back-pressure to the host. I accept that over PCIe this is not a problem, but the purpose of designing a general IP is that it's usable in any situation. If this were in an AXI-based SoC, the host processor could send many reads/writes and overload TX Fifo, which wouldn't accept AXI writes and possibly crash the host 
 
- There's no way of tracking network credits. A solution to above would be to have the host track network credits and have more entries in the TX FIFO than the credits initialized on the host. This would also solve   (#253) and provide fencing for free. I didn't put this in my original review but now seems relevant to this PR.

leonardxiang:
Thanks all!

For the 1st concern, I agree that we should avoid letting the AXI master (PCIe or Soc) overload the TX fifo, even though it is not ready to accept more incoming request. Like professor said, the host can hold a cached copy of the tx-fifo vacancy register.

For the 2nd concern, I think what @drichmond said was, how to track the network credits?
There are two kinds of credits that are tracked;
1) host write request to the manycore. This is tracked by the manycore_endpoint_standard, and we used it to fence the Tx-FIFO
2) host read request to the manycore. This can be tracked in either the host or the manycore_link_to_axil module:
image

leonardxiang:
To be clear,

About the 1st concern, not overflowing the TX FIFO. The host should be able to track its credits, which is set not larger than the TX FIFO.

About the 2nd concern, how does the host track the returning credits?
1) For read, just increment the credit counter after fetching the returned packet.
2) For write, the host should check that both the TX-FIFO's vacancy and the endpoint's out_credits have resumed to their maximum value. This indicates that all writes are finished, so that the host can reset its credits to the maximum.
Like @drichmond said, This would also solve (#253) and provide fencing for free.

drichmond:
This is a bit tricky. Maybe, as a solution, when sending data:

1. Send until local copy of vacancy is 0. 
2. Update local copy of vacancies. 
3. Repeat until complete. 

An optimization might be to update the local vacancy copy before starting a transfer if the local copy has less than half of the maximum remaining. 

Then when fencing, we can just iterate until credits returns to maximum.

drichmond:
@leonardxiang, what's the status on this?

leonardxiang:
> @leonardxiang, what's the status on this?

Hi, Dustin
I guess this is nearly completed.
According to the latest concerns, the host request flow control method has been changed:
- Host now tracks a cached version of the credits when sending.
- The write fence function has been implemented.
- Flow control parameters are added in the ROM.
- Verified in cosim and FPGA.

drichmond:
OK, so we believe there's no way that the host can send too many packets?

leonardxiang:
Yes, by using of the function **hb_mc_manycore_host_request_fence()**
Note that the flow control for host sending packets is above the API of hb_mc_manycore_packet_tx()

drichmond:
I promise that I'll get to this tomorrow

Makefile.machine.include
```diff
@@ -37,12 +37,16 @@ BSG_MACHINE_MAX_EPA_WIDTH             = 28
 BSG_MACHINE_DRAM_SIZE_WORDS           = 536870912
 BSG_MACHINE_DRAM_BANK_SIZE_WORDS      = 134217728
 
+BSG_MACHINE_IO_REMOTE_LOAD_CAP        = 32
+BSG_MACHINE_IO_HOST_CREDITS_CAP       = 256
+BSG_MACHINE_IO_EP_MAX_OUT_CREDITS     = 16
+
```
drichmond: Can you add these to all of the machine files, and add a sentence describing what they do?

hardware/bsg_manycore_endpoint_to_fifos.v
```diff
@@ -0,0 +1,256 @@
...
+  bsg_mcl_response_s mc_rsp_lo_cast  ;
+  assign host_req_li_cast = host_req_i;
+  assign mc_rsp_o         = mc_rsp_lo_cast;
+
+
+  // manycore as master
+  bsg_mcl_request_s mc_req_lo_cast;
+  assign mc_req_o = mc_req_lo_cast;
+  // assign host_rsp_li_cast = host_rsp_i;
+  // bsg_mcl_response_s host_rsp_li_cast;
```
drichmond: If these aren't relevant, then we can remove them

hardware/bsg_manycore_endpoint_to_fifos.v
```diff
@@ -0,0 +1,256 @@
...
+  // host request to manycore
+  // -------------------------
+  assign endpoint_out_v_li = ~(out_credits_o == 0) & host_req_v_i;
+  assign host_req_ready_o  = ~(out_credits_o == 0) & endpoint_out_ready_lo;
+
+  assign endpoint_out_packet_li.addr       = addr_width_p'(host_req_li_cast.addr);
+  assign endpoint_out_packet_li.op         = bsg_manycore_packet_op_e'(host_req_li_cast.op);
+  assign endpoint_out_packet_li.op_ex      = bsg_manycore_packet_op_ex_u'(host_req_li_cast.op_ex);
+  assign endpoint_out_packet_li.reg_id     = bsg_manycore_reg_id_width_gp'(host_req_li_cast.reg_id);
+  // assign endpoint_out_packet_li.payload    = bsg_manycore_packet_payload_u'(host_req_li_cast.payload);
```
drichmond: OK, I'll stop commenting on commented lines. But they should be removed if they aren't relevant.

An example of being relevant - is if we anticipate having the host respond to manycore packets in the future. Those ports are somewhat useful to keep around in case a future developer has questions.

Anyway, your discretion.

hardware/bsg_manycore_endpoint_to_fifos.v
```diff
@@ -0,0 +1,256 @@
...
+    .link_sif_o           (link_sif_o            ),
+
+    // manycore packet -> fifo
+    .in_v_o               (endpoint_in_v_lo      ),
+    .in_yumi_i            (endpoint_in_yumi_li   ),
+    .in_data_o            (endpoint_in_data_lo   ),
+    .in_mask_o            (endpoint_in_mask_lo   ),
+    .in_addr_o            (endpoint_in_addr_lo   ),
+    .in_we_o              (endpoint_in_we_lo     ),
+    .in_load_info_o       (                      ), // not used since remote load is not supported on host interface.
```
drichmond: I think I know what you mean here, but I'm an expert.

What do you mean by "not used since the remove load is not supported on the host interface". That the host can't respond to read requests / the manycore can't issue read requests to the host?

hardware/bsg_manycore_link_to_axil.v
```diff
@@ -0,0 +1,387 @@
...
+// ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+// SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+
+/*
+*  bsg_manycore_link_to_axil.v
+*
+*      AXIL memory-mapped (HOST side)   <->   manycore endpoint standard (MC side)
+*
```
drichmond: This is a pretty small comment for what is a very large (and important?) module. It's fine if you copy some of the description from the google doc, but I think its useful to have a good description at the top of an important module

hardware/bsg_manycore_link_to_axil.v
```diff
@@ -0,0 +1,387 @@
...
+  logic                         tx_wen_li   ;
+  logic                         tx_wready_lo;
+
+  // --------------------------------------------
+  // axil write channels
+  // --------------------------------------------
+  // typical dependency:
+  // awready : awvalid or wvalid or None
+  // wready : awvalid or wvalid or None
+  // bvalid : after the cycle of wvalid & wready
```
drichmond: What is meant by "typical dependency"? 

My understanding from the axi specification is that awready is unrelated to wvalid, and wready is unrelated to awvalid. I'm just confused what is being conveyed here.

hardware/bsg_manycore_link_to_axil.v
```diff
@@ -0,0 +1,387 @@
...
+
+  always_comb begin
+    awready_lo = 1'b0;
+    wready_lo  = 1'b1;
+
+    tx_wen_li   = 1'b0;
+    tx_wdata_li = '0;
+
+    bvalid_n = bvalid_r;
+    bresp_lo = 2'b00; // always OKAY even writing to the undefined address
```
drichmond: It'd be clearer if 2'b00 were a macro. Just a thought

hardware/bsg_manycore_link_to_axil.v
```diff
@@ -0,0 +1,387 @@
...
+  // 4. rom
+  logic [axil_addr_width_p-1:0] rx_rom_addr_li;
+  logic [axil_data_width_p-1:0] rx_rom_data_lo;
+
+  // --------------------------------------------
+  // axil read channels
+  // --------------------------------------------
+  // typical dependency:
+  // arready : arvalid or None
+  // rvalid : arvalid & arready
```
drichmond: See comment above

hardware/bsg_manycore_link_to_axil.v
```diff
@@ -0,0 +1,387 @@
...
+  wire is_read_rom           = (axil_araddr_i >= mcl_rom_base_addr_gp) &&
+                               (axil_araddr_i < mcl_rom_base_addr_gp + (1<<$clog2(rom_els_p*rom_width_p/8)));
+
+  always_comb begin
+
+    arready_lo = 1'b0;
+    rdata_n = rdata_r;
+    rvalid_n = rvalid_r;
+
+    rresp_lo = 2'b00; // always OKAY even reading from the undefined address
```
drichmond: See comment above

hardware/bsg_manycore_link_to_axil.v
```diff
@@ -0,0 +1,387 @@
...
+      rvalid_r <= rvalid_n;
+    end
+  end
+
+  // ----------------------------
+  // bladerunner rom
+  // ----------------------------
+  // The rom not necessarily in the mcl, so we put its parameters in other package.
+  localparam rom_addr_width_lp = `BSG_SAFE_CLOG2(rom_els_p); // TODO: _gp?
+
```
drichmond: Still TODO?

hardware/bsg_manycore_link_to_axil.v
```diff
@@ -0,0 +1,387 @@
...
+    .addr_i(br_rom_addr_li),
+    .data_o(br_rom_data_lo)
+  );
+
+
+  // --------------------------------------------
+  // axil fifo data stream
+  // --------------------------------------------
+
+  // host ---packet---> mc
```
drichmond: These are great, very useful.

hardware/bsg_manycore_link_to_axil_pkg.v
```diff
@@ -0,0 +1,80 @@
+`ifndef BSG_MANYCORE_LINK_TO_AXIL_PKG_V
+`define BSG_MANYCORE_LINK_TO_AXIL_PKG_V
+
+
+// this macro inherited the parameters of bsg_manycore_pkg.v
+// it slices the manycore packet struct into fields of multiple bytes, and the size of new packet is N words
```
drichmond: This comment isn't super clear. What is it trying to say?

hardware/bsg_mcl_axil_fifos_master.v
```diff
@@ -0,0 +1,222 @@
...
+// WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+// DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR
+// ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+// (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+// LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+// ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+// SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+
```
drichmond: This module is missing a descriptive comment

hardware/bsg_mcl_axil_fifos_master.v
```diff
@@ -0,0 +1,222 @@
...
+        fifo_width_p, axil_data_width_p);
+  end
+  // synopsys translate_on
+
+  localparam sipo_cnt_width_lp = $clog2(sipo_els_lp+1);
+
+
+  // --------------------------------------------------------
+  //                     Data Path Fwd
+  // --------------------------------------------------------
```
drichmond: What does fwd mean in this context?

hardware/bsg_mcl_axil_fifos_slave.v
```diff
@@ -0,0 +1,113 @@
...
+  ,output                         mc_req_ready_o
+  ,output [ pkt_cnt_width_lp-1:0] mc_req_words_o
+);
+
+  // synopsys translate_off
+  initial begin
+    assert (piso_els_lp * axil_data_width_p == fifo_width_p)
+      else $fatal("BSG_ERROR][%m]: manycore link fifo width %d is not multiple of axil data width %d.",
+        fifo_width_p, axil_data_width_p);
+  end
```
drichmond: 👍 awesome

hardware/cl_manycore.sv
```diff
@@ -31,6 +31,8 @@
 
 `include "bsg_bladerunner_rom_pkg.vh"
 
+// `undef COSIM
+
```
drichmond: Might want to remove this.

hardware/cl_manycore.sv
```diff
@@ -351,9 +353,6 @@ module cl_manycore
       );
 
 
-
-`ifdef COSIM
-
```
drichmond: I think, this should still be here. print_stat signals aren't synthesizable?

hardware/cl_manycore_pkg.v
```diff
@@ -40,7 +40,7 @@ package cl_manycore_pkg;
   parameter axi_burst_len_p = 1;
 
   // the max number of outstanding requests from the host endpoint to the manycore
-  parameter max_out_credits_p = 16;
+  parameter max_out_credits_p = `CL_MANYCORE_IO_EP_MAX_OUT_CREDITS;
```
drichmond: 👍

hardware/hardware.mk
```diff
@@ -151,6 +154,9 @@ $(HARDWARE_PATH)/bsg_bladerunner_configuration.rom: $(BSG_MACHINE_PATH)/Makefile
 	@echo $(call dec2bin,$(CL_MANYCORE_VCACHE_BLOCK_SIZE_WORDS))  >> $@.temp
 	@echo $(call dec2bin,$(CL_MANYCORE_VCACHE_STRIPE_SIZE_WORDS)) >> $@.temp
 	@echo $(call dec2bin,$(CL_MANYCORE_VCACHE_MISS_FIFO_ELS)) >> $@.temp
+	@echo $(call dec2bin,$(CL_MANYCORE_IO_REMOTE_LOAD_CAP)) >> $@.temp
+	@echo $(call dec2bin,$(CL_MANYCORE_IO_HOST_CREDITS_CAP)) >> $@.temp
+	@echo $(call dec2bin,$(CL_MANYCORE_IO_EP_MAX_OUT_CREDITS)) >> $@.temp
```
drichmond: Can you add a check for these in test_rom?

libraries/bsg_manycore_config.cpp
```diff
@@ -135,13 +135,19 @@ int hb_mc_config_init(const hb_mc_config_raw_t raw[HB_MC_CONFIG_MAX],
...
                 return HB_MC_INVALID;
         }
         config->vcache_stripe_words = idx;
 
+        config->io_remote_load_cap = raw[HB_MC_CONFIG_IO_REMOTE_LOAD_CAP];
+
+        config->io_host_credits_cap = raw[HB_MC_CONFIG_IO_HOST_CREDITS_CAP];
+
+        config->io_endpoint_max_out_credits = raw[HB_MC_CONFIG_IO_EP_MAX_OUT_CREDITS];
+
```
drichmond: Can you check for invalid values here?

regression/library/test_rom.c
```diff
@@ -57,20 +57,38 @@ int test_rom (int argc, char **argv) {
...
+            }
         }
 
+        if (!matched) {
+            bsg_pr_test_err("Incorrect number of out credits. "
+                            "Got: %d, expected %d\n", result, expected);
+            bsg_pr_test_err("Have you programed your FPGA"
+                            " (fpga-load-local-image)\n");
+            fail = 1;
+            goto cleanup; // Considered harmful
```
drichmond: 👀

hardware/hardware.mk
drichmond: I see you've done some checking in test_rom. Maybe that is sufficient. Let me know if it is

hardware/cl_manycore.sv
leonardxiang: Yes, you are right, I should undo this after testing.

libraries/bsg_manycore_config.h
```diff
@@ -350,6 +361,47 @@ extern "C" {
...
+        {
+                return cfg->io_host_credits_cap;
+        }
+
+        /**
+         * Return the threshold for host to update its cached credits.
+         * @param[in] cfg A configuration initialized from the manycore ROM.
+         * @return the threshold for host to update its cached credits.
+         */
+        static inline uint32_t hb_mc_config_get_io_credit_upate_threshold(const hb_mc_config_t *cfg)
```
drichmond: Should this be `update`?

libraries/bsg_manycore.cpp
```diff
@@ -1526,17 +1421,34 @@ int hb_mc_manycore_write_mem(hb_mc_manycore_t *mc, const hb_mc_npa_t *npa,
...
                 hb_mc_npa_set_epa(&addr, hb_mc_npa_get_epa(&addr) + 4);
         }
 
+        err = hb_mc_manycore_host_request_fence(mc);
+        if (err != HB_MC_SUCCESS)
+                return err;
+
+        err = hb_mc_manycore_host_requests_init(mc);
+        if (err != HB_MC_SUCCESS)
+                return err;
```
drichmond: Why do you call init here? Shouldn't it be 0 after a fence?

libraries/bsg_manycore_config.cpp
```diff
@@ -135,13 +135,28 @@ int hb_mc_config_init(const hb_mc_config_raw_t raw[HB_MC_CONFIG_MAX],
...
+                return HB_MC_INVALID;
+        }
+        config->io_remote_load_cap = idx;
+
+
+        /* the flowing flow control parameters are checked at test_rom*/
+
+        config->io_endpoint_max_out_credits = raw[HB_MC_CONFIG_IO_EP_MAX_OUT_CREDITS];
+        config->io_host_credits_cap = raw[HB_MC_CONFIG_IO_HOST_CREDITS_CAP];
+
```
drichmond: yeah, but test_rom isn't run every time. You should check that they're sensible values here -- i.e non zero, not 0xFFFFFFFF, and less than something else.

libraries/bsg_manycore_config.h
```diff
@@ -59,7 +59,21 @@ extern "C" {
...
+
+typedef enum __hb_mc_endpoint_out_credits_limit_t {
+        HB_MC_EP_OUT_CREDITS_MIN = 1,
+        HB_MC_EP_OUT_CREDITS_MAX = 64
+} __hb_mc_endpoint_out_credits_limit_t;
+
+typedef enum __hb_mc_host_credits_limit_t {
+        HB_MC_HOST_CREDITS_MIN = 1,
+        HB_MC_HOST_CREDITS_MAX = 512
+} __hb_mc_host_credits_limit_t;
```
drichmond: Enums are overkill. Can you just make them macros?

drichmond:
@gaozihou Do you have time to review the hardware-side of this PR?

hardware/bsg_manycore_link_to_axil_pkg.v
```diff
@@ -0,0 +1,83 @@
...
+  import cl_manycore_pkg::x_cord_width_p;
+  import cl_manycore_pkg::y_cord_width_p;
+
+  parameter  mcl_fifo_width_gp    = 128;
+  parameter  mcl_read_credits_gp  = 256;
+  parameter  mcl_edpt_fifo_els_gp = 4  ;
+  localparam axil_data_width_gp   = 32 ;
+  localparam axil_addr_width_gp   = 32 ;
+
+  localparam axil_resp_OKAY_gp = 2'b00;
```
drichmond: Do we prefer parameters instead of macros in basejump? (Asking for my own edification)

hardware/bsg_manycore_link_to_axil_pkg.v
drichmond: Did you remove this comment? @leonardxiang

leonardxiang: I replaced it with this:
// To ease the programming, we slices the fields in the bsg_manycore_packet_s
// struct, and cast the fileds into multiple of bytes.
Hope it is clear to others.

hardware/bsg_manycore_link_to_axil_pkg.v
dpetrisko: Yep

libraries/bsg_manycore_request_packet.h
leonardxiang: Previously, it was named load_id. But after the AMO change, I guess it is changed to reg_id.
I don't know why, maybe @mrutt92 can explain.

hardware/bsg_manycore_link_to_axil_pkg.v
drichmond: Cool

hardware/bsg_mcl_axil_fifos_master.v
```diff
@@ -0,0 +1,229 @@
...
+    .reset_i(reset_i        ),
+    .v_i    (w_v_i          ),
+    .ready_o(w_ready_o      ),
+    .data_i (w_data_i       ),
+    .v_o    (req_buf_v_lo   ),
+    .data_o (req_buf_data_lo),
+    .yumi_i (req_buf_yumi_li)
+  );
+
+  bsg_serial_in_parallel_out #(
```
gaozihou: bsg_serial_in_parallel_out_full should also work: https://github.com/bespoke-silicon-group/basejump_stl/blob/master/bsg_dataflow/bsg_serial_in_parallel_out_full.v

bsg_sipof is implemented with buffer and mux, while bsg_sipo is implemented with shifting registers. The number of words extracted from bsg_siop can vary dynamically, but in this module sipo_els_lp is fixed, so we might want to use the light-weight bsg_sipof?

leonardxiang: Thanks Paul!
In terms of fixed and large sipo_els_lp, The **sipof** implementation is definitely a good choice.
But considering the possible overhead from the **round robin** and the **two_fifo** instantiated in **sipo_full**, I am not sure if it is light than **sipo** in this use case.
It would be also good to have a PPA comparison of these two modules.
I guess that the power consumption of shifting registers will be higher?

gaozihou: In SAED90:

`bsg_serial_in_parallel_out_width_p32_els_p4_1:   total area 9034.44   comb area 5675.21   seq area 3359.23`
`bsg_serial_in_parallel_out_full_width_p32_els_p4_1:   total area 6688.97   comb area 2483.71   seq area 4205.26`

For timing, bsg_sipo has larger positive slack than bsg_sipof after synth. However, looking into bsg_sipo, data_i is bypassed to data_o directly, these paths are not considered in post-synth timing analysis. 

Switching power of bsg_sipo is 409.4351 uW, bsg_sipof is 146.6041 uW.

leonardxiang: WOW, so much improvement! It is indeed better to use sipof. Thanks for this stats Paul.
So, can I get the conclusion that sipof shall always have better PPA if the els_p is fixed?

gaozihou: Weird thing is the internal power above, seems that internal power of bsg_sipof is worse than bsg_sipo, not sure about what this implies...

drichmond: I'm a bit behind here. Did we chose to go with sipof? sipo has shift registers and sipof does not?

In my experience, shift registers are slower to simulate and synthesize. If I've read this correctly, this is a good choice.

On an unrelated note, I'd also caution against applying PPA regression for ASIC to FPGA synthesis results. 99% of the time they're probably correlated...

Issue #505, opened 2020-02-11
hb_mc_device_init should allow users to initialize less tiles than the whole array


drichmond:
The default behavior should be all the tiles, but if users want their simulation init to be faster, there should be a method to initialize a subset. 

Borna's excellent experiments today demonstrated that this can reduce cosimulation time by 5x for 1x1 tile groups (which I think is substantial)

@mrutt92 Any thoughts on default arguments?

mrutt92:
Not sure. Isn't the dimensionality set by kernel_enqueue?

mrutt92:
I guess program_init() primes the caches on every node. Seems like that could slow things way down if you're only using a few cores.

tommydcjung:
I would suggest for the memory models where we can add $readmemb, create a bit vector file for each bank, and completely bypass host interface. That should be +10x improvement.

mrutt92:
Yeah and that'll probably be addressing the real issue more than fussing with the default behavior of CUDA.

drichmond:
What tommy is suggesting only works for external memory, right?

mrutt92:
No, I think what Tommy is suggesting works for any of the RAMs (I hope).
I'm already working on a solution to DRAM IO which is a similar approach (except I've moved data store into C++ and I'm using memcpy instead of readmemb).

drichmond:
Since we're switching to C++ models for DRAM, that solution sounds reasonable to me. 

For the kernels that this Issue affects, DRAM overhead is small. The code Borna and I worked on earlier took 100 seconds to execute with full array initialization, vs 19 to execute with single tile-initialization.

taylor-bsg:
Maybe we should run the profiler on hb_mc_device_init ?   ;-)

drichmond:
But we know what the issue is - the runtime has to initialize the instruction cache of every single tile. If your program only uses one tile this is a huge overhead.

taylor-bsg:
Initializing the icache is not an inherently slow operation. We should see why it is slow.

drichmond:
It is slow because IO is slow. For every ICache we have to inititialize every entry with program data. That means every tile requires 1K writes. For a 4x4 manycore array, that's 16K 4-byte writes, or 64K of data being uploaded to the device before a program can run. 

Why do people think this is a controversial change? Unlike a backdoor to the device, it isn't introducing any magic. We've demonstrated that there's a 5x speedup and it requires a simple change to the API that will only affect a few users.

[...]

PR #506, opened 2020-02-12
Acceleration of simulation IO with back door to DRAM data store

mrutt92:
Relies on https://github.com/bespoke-silicon-group/basejump_stl/pull/236
Relies on https://github.com/bespoke-silicon-group/bsg_manycore/pull/232

This is still a WIP. To really work we need a way to notify caches that they should invalidate all their data. Writing works pretty well for most of our CUDA applications by first of the fact that we typically don't access the memory locations we write to before calling memcpy().

Reading the results out is totally broken ATM because of the cache issue which is why `#define DRAM_HACK_READ` is commented out.


mrutt92:
BTW - I haven't measured the speedup yet but I have a rough idea, from observing, that it's crazy good.

drichmond:
Great! As part of the experiments, can you run a kernel which is small, a 1x1 tile group, and has very little DRAM data?

Borna and I can provide one if this works seamlessly with baseline

mrutt92:
How much is "very little?"

and does it actually need to run anything?

mrutt92:
and also why? the whole point is to speedup IO with large input.

drichmond:
512 words, ish. The vector-vector add example that I referred to in #505 is adding two 128 element vectors.

drichmond:
Yes, this is exactly my point in #505

mrutt92:
I mean... it's not going to provide much of a speedup for an application with no DRAM data - the problem of initializing a bunch of tiles is not the problem that this PR attempts to solve.

This PR addresses a problem we have when we need to move a large amount (like megabytes) of input data into the device's DRAM and it takes hour just to perform IO but likely not much time to actually do the compute.

The issue you raise in #505 is an important one but it's not addressed by this PR (nor is it meant to be).

mrutt92:
Admittedly this PR is addressing an issue that those of us who are working on graph applications are running into, but maybe not so much for the ML and linear algebra folks.

drichmond:
I have no beef with this PR, I think it's a great PR. I look forward to switching to the C++ models and this will really make our cosimulation sing.

drichmond:
Both issues are addressing an important issue that Cornell raised today - why does it take so long to do initialization when your ROI is only 5000 cycles

drichmond:
So timely, and great.

taylor-bsg:
A couple of different alternatives to address the incoherency issue:

a. add a backdoor function to the vcache implementations that magically
updates any bytes in the cache. If the line is not dirty, it could just
invalidate the line. Messy, because there are transient states.

b. Add a host routine that issues a stream of tagla’s to read the tags out
of the caches. Then issue a stream of parallel tagfl, tagsw instructions to
invalidate any potential conflicts. This could be useful for our silicon
where we are hosting the DRAM on an FPGA or PC via ASIC memory backdoor
interface. We could do direct writes to the DRAM without having to go over
the I/O link to the ASIC.



M

mrutt92:
> b. Add a host routine that issues a stream of tagla’s to read the tags out
of the caches. Then issue a stream of parallel tagfl, tagsw instructions to
invalidate any potential conflicts. This could be useful for our silicon
where we are hosting the DRAM on an FPGA or PC via ASIC memory backdoor
interface. We could do direct writes to the DRAM without having to go over
the I/O link to the ASIC.

@taylor-bsg yeah - I was thinking of a solution more in the vein of (b).

drichmond:
Weird idea - but could we have the manycore array somehow clear the tags? Then it could be done in parallel when the cores boot?

mrutt92:
I was thinking that instead of TAGFL, TAGST, TAGL we might want to use AFL before DRAM reads and AINV after DRAM writes?

mrutt92:
Oh wait, I see that TAGL and TAGST are already be supported? Just set the high-order EPA bit to 1?

mrutt92:
The problem that I see with going down this road (by which I mean relying on TGSTW, TAGLA, and flushing in *either* hardware or software) is it would mean doing a fair amount of IO in simulation just to deal with the cache. N-way associativity means you have to either do up to N-loads to find the tag you want to invalidate and then do a store *OR*  just do. N-stores to invalidate the whole set. Flushing would be painful for similar reasons. 

None of this seems like a big deal for actual hardware (where, frankly, the problem addressed by this PR is completely moot) but this cost in simulation would eat away at the gains from this PR and it would not scale with associativity. In fact, if associativity is greater than a line size in words, this actually makes IO worse because you would ultimately be sending more packets onto the network than if you just performed IO directly by the host-interface. You could keep track of sets that have already been invalidated/flushed and not repeat yourself, and then your worst case is the size of the cache, but that's still a bummer.

That's why I'm wondering if we shouldn't just add support for AINV and AFL somehow. It's lame to add hardware for the reason of simulating faster though...

vb000:
> a. add a backdoor function to the vcache implementations that magically updates any bytes in the cache. If the line is not dirty, it could just invalidate the line. Messy, because there are transient states.

Since the backdoor is solely for simulation, isn't this a better choice? If we have a non-synthesizable sv function in vcache, which could be invoked to initializes cache-lines with the same data being loaded to the dram while maintaining the state (dirty bit), transient states are undisturbed, right?

taylor-bsg:
Max, afl is for small chunks of data and tagfl is for large ranges of
address space. Both are useful generally for doing software coherence, but
in our use case afl would be proportionally the same cost to write the data
into the cache directly over the network than use afl. The beauty of tagfl
is I can flush a gigabyte of data just by scanning 1024 tags.

M

mrutt92:
Well... I ended up going with more or less @taylor-bsg's suggestion.

drichmond:
Replaced by #521

PR #507, opened 2020-02-18
fixes the initialization of icache

mrutt92:
Before: the tag data in the MSBs was set to 0 always
Now: we set the tag data do the MSBs of the pc (we should have been doing this the whole time).

### DIFF
--- a/libraries/bsg_manycore_loader.cpp
+++ b/libraries/bsg_manycore_loader.cpp
@@ -364,6 +364,8 @@ static int hb_mc_loader_load_tile_icache(hb_mc_manycore_t *mc,
                    hb_mc_npa_get_epa(&icache_npa)
                    );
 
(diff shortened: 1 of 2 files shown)


mrutt92:
there's no issue with the code as written - I had been confusing myself with a stale binary.

drichmond:
Stale manycore binary?

mrutt92:
Yeah - I was seeing icache misses that I didn't think should be there but it turned out that when you run `make <test>.log` from bsg_replicant it doesn't automatically rebuild `main.riscv`

drichmond:
Yeah, this is a known issue. I've tried to solve it several times unsuccessfully.

drichmond:
(fwiw, solved in baseline)

mrutt92:
not trying to point fingers - just explaining why I closed the PR for posterity.

drichmond:
No worries, didn't feel like you were. Just trying to understand what stale binary meant. 

It's been a frustration for me too, obviously.

PR #508, opened 2020-02-19
Option to turn off waveforms

vb000:
- Adds makefile flag to toggle waveform dumps.
- Turn off waveform dumps by default. Waveforms can be enabled by setting `WAVES=1`.
- AWS testbench implements a weird construct which requires debug hooks during compilation. So this change uses the runtime sv option it provides called `+NO_WAVES`.
- This fixes the annoying problem of not being able to kill VCS with Ctrl-C :)
- ~2x speedup in cosim!

### DIFF
--- a/testbenches/cosimulation.mk
+++ b/testbenches/cosimulation.mk
@@ -76,13 +76,17 @@ help:
 	@echo "      <test_name>: Build the cosimulation binary for a specific"
 	@echo "             test"
 	@echo "      <test_name.log>: Run a specific cosimulation test and "
-	@echo "             generate the log file"
+	@echo "             generate the log file. Set WAVES=1 for waveform dump."
 	@echo "      <test_name.vanilla.log>: Run a specific cosimulation test and "
 	@echo "             generate the vanilla log file"
 	@echo "      clean: Remove all subdirectory-specific outputs"
 
 $(REGRESSION_TESTS): %: $(EXEC_PATH)/%
 test_loader: %: $(EXEC_PATH)/%
+
+ifeq ($(WAVES),)
+$(EXEC_PATH)/%.log: SIM_ARGS += +NO_WAVES
+endif
 
 VPD_RULES = $(addsuffix .vpd,$(REGRESSION_TESTS))
 $(VPD_RULES): SIM_ARGS +=+trace


mrutt92:
I was under the impression that something like this already existed? @drichmond isn't there a debug flag?

vb000:
Oh! Without mods to aws infrastructure, waveforms can only be disable with `NO_WAVES`. I couldn't find it when I grepped for it. May be there is, @drichmond?

Either way, I think it's a standard practice to disable waveforms by default, so that user has to be explicit to enable them...

mrutt92:
I agree - I just think it would be confusing to have a DEBUG flag and WAVE flag. Would you need to set both to get a waveform? Because that would be annoying.

vb000:
When I tested with this pr mods, `WAVES=1` writes the vcd file with no other flag needed.

drichmond:
I'm open to suggestions on this. Should waveforms be generated by default? Yes, there is DEBUG (I don't recall what is set by default). 

In baseline we generate everything by default. I got tired of everyone asking "how to do I generate X"

I like this PR, but I'm having a think about how to do it. Let me think about it some.

taylor-bsg:
I think it just comes down to the performance difference?

drichmond:
Yep. Which is why some form of this will get accepted at some point. 

But the reality is that regression shouldn't be producing waveforms anyway. I wrote it that way, so something has changed and I need time to go debug it, and then fix it.

dpetrisko:
My 2c. Default option having debugging is more useful, because most manual tests you run are failing tests that you want to debug. Regression scripts can easily override default options without having to “think” about it each time.

mrutt92:
+ to @dpetrisko

drichmond:
To be clear - if you run a single test, you want the vpd file to be generated by default?

drichmond:
Vs, if you run regression no vpd files are generated

taylor-bsg:
Interesting point by Dan. In BlackParrot, the common case is you are
debugging the processor on a piece of code that is supposed to work. In
HammerBlade the common case is you are debugging the program running on
HammerBlade. So probably tracing but not VCD makes sense for HB (assuming
here that tracing has little impact on execution time.)

mrutt92:
Honestly... it really depends on the person. I feel like we're at a point with HB where the people working on this are 50/50 on whether their working on the RTL (in which case you probably want the VPD) or your working on the software and you're more interested in the profiling information.

drichmond:
@taylor-bsg To my knowledge, nobody has profiled the tracer. I've asked myself the same question several times. 

A proposal, for everyone: 

The way I'd switch to doing it, if I hop in and do it, is that I'd create a specific target for the .vpd file. If you don't ask for that .vpd file, you don't get it. Likewise, if you don't ask for it, you don't pay the performance penalty. Ditto for profiling and tracing. This is the way make is designed to work, though it may result in multiple compilation runes. I tried this out in baseline recently and I'm actually kinda happy with the results.

Reasoning: 

We used to have a lot more Makefile "arguments" in this repository and the cross-product of different ones became confusing. Like, what happens when you run with TURBO and DEBUG. Or AXI_MEMORY_MODEL and DEBUG. Slowly those have become either explicit, or machine-configuration variables so that the code is easier to document. This evolution has also come as I've learned more about the tools (vcs) and had a lot of frustration with long-lived makefile "arguments" that influence subsequent simulation (Yes, it's weird, but a couple did). 

For example: if you really want absolute bleeding-edge performance (in simulation, hah) you need to recompile the synopsys-intermediate-verilog-representation without the Verilog VCS macro. This prevents the .vpd file from being generated because of some odd code in the AWS infrastructure (not ours) and allows you to remove the -debug_pp flag. If I recall correctly this got a measurable performance increase over simulation that doesn't have VPD files being generated (30% or something), and that's in addition to the 2x that Bandhav (and myself) have seen in the past.

I can write rules avoid all this really funky behavior because I can encode the explicit dependencies better now that I've learned how VCS works. 

(Aside: the complexity for profiling and tracing is that they generate generic names: vanilla_stats.csv, vanilla_operation_trace.csv, etc. They get overwritten when you run multiple tests. Also solved in baseline).

drichmond:
TL;DR - VCS is a painful tool to write make rules for

vb000:
I personally felt most it's best to start RTL debugging with the trace. And waveform is only needed when developing something from scratch or the bug pretty nasty one.

drichmond:
A bit of a divergent question 

Should traces/statistics be generated by default?

mrutt92:
So here's the thing... nothing is more triggering than running a long simulation only to realize the trace info you needed wasn't generated.

mrutt92:
But yeah, idk - does it affect performance?

vb000:
Traces and stats on by default and waveforms off by default? Waveforms are slowest to write to, and are not usually necessary to know what went wrong if traces are available...

I think the only option that's strictly required to be added to current infrastructure is the ability to turn off waveforms. Default or not? Above is my take, but it's a matter of choice.

dpetrisko:
>Waveforms are slowest to write to
@vb000 Have you measured this? I've observed the opposite in BP for some levels of trace verbosity.  I'd believe it, but I'm curious where the inflection point is

vb000:
@dpetrisko No, that's an empirical observation on bsg_manycore.

There are python scripts to post-process raw cycle accurate trace. So, traces should definitely contain significantly less number of signals compared to vpd. I heard the vpd format is well optimized for size, but I don't think they are faster than traces in manycore's case.

vb000:
I did a quick run on test_fib:

waveform on: 14.94s
vanilla.log enabled: 11.37s
no logs or waveforms: 8.59s

All wall clock times and compilation excluded. It's a really small test, but the disparity only grows as test gets bigger.

dpetrisko:
Thanks Bandhav! What kind of info is printed in vanilla.log, pc/instruction traces and cycle stalls?

vb000:
matrix_mul:

only waveform on: 2m32.8s
only vanilla.log enabled: 2m18s
no logs or waveforms: 1m26s

@dpetrisko's right... traces are pretty close to vpds!

drichmond:
Does test_fib only run on one core? 

I suppose that might not matter - since I think trace still runs on all cores 

Also curious about the effect of debug_pp. I have some experiments coded up that I can run tomorrow and hopefully get results for all the regression tests

vb000:
> Thanks Bandhav! What kind of info is printed in vanilla.log, pc/instruction traces and cycle stalls?

Time, tile-coordinates, int pc instr reg-writes, stalls, float pc instr reg-writes, branch-target, remote-loads.

drichmond:
Another experiment (after this one) is that we should remove the `[INFO][VCORE] making/breaking reservation statements` in the manycore and run timing experiments. We're printing out thousands of those lines per simulation run in some CUDA kernels

vb000:
Dustin's on travel... posting on his behalf:

```
TEST                                         debug_pp_vpd_trace debug_pp_vpd_no_trace debug_pp_no_vpd_no_trace no_debug_pp_no_vpd_trace no_debug_pp_no_vpd_no_trace 
./cuda/test_barrier.log                      50.250 seconds     48.760 seconds        44.690 seconds           38.220 seconds           36.910 seconds              
./cuda/test_binary_load_buffer.log           44.570 seconds     44.630 seconds        43.870 seconds           39.270 seconds           34.020 seconds              
./cuda/test_conv1d.log                       57.550 seconds     89.750 seconds        44.470 seconds           114.510 seconds          42.680 seconds              
./cuda/test_conv2d.log                       52.550 seconds     46.780 seconds        46.030 seconds           47.030 seconds           45.030 seconds              
./cuda/test_device_memcpy.log                52.620 seconds     47.750 seconds        47.780 seconds           45.270 seconds           37.330 seconds              
./cuda/test_device_memset.log                46.590 seconds     45.930 seconds        44.180 seconds           45.830 seconds           37.240 seconds              
./cuda/test_dram_device_allocated.log        76.600 seconds     59.410 seconds        57.150 seconds           66.890 seconds           45.310 seconds              
./cuda/test_dram_host_allocated.log          48.380 seconds     49.280 seconds        46.280 seconds           38.490 seconds           38.650 seconds              
./cuda/test_dram_load_store.log              93.240 seconds     69.960 seconds        69.770 seconds           90.860 seconds           56.030 seconds              
./cuda/test_empty.log                        42.070 seconds     43.190 seconds        43.630 seconds           35.020 seconds           34.000 seconds              
./cuda/test_empty_parallel.log               50.650 seconds     45.070 seconds        42.610 seconds           41.540 seconds           35.020 seconds              
./cuda/test_float_all_ops.log                222.720 seconds    114.030 seconds       117.790 seconds          215.320 seconds          94.340 seconds              
./cuda/test_float_matrix_mul.log             144.730 seconds    87.190 seconds        88.230 seconds           140.290 seconds          71.890 seconds              
./cuda/test_float_matrix_mul_shared_mem.log  345.240 seconds    176.690 seconds       175.540 seconds          310.020 seconds          128.200 seconds             
./cuda/test_float_vec_add.log                77.860 seconds     59.060 seconds        58.370 seconds           74.060 seconds           48.270 seconds              
./cuda/test_float_vec_add_shared_mem.log     103.580 seconds    67.730 seconds        63.430 seconds           101.140 seconds          58.720 seconds              
./cuda/test_float_vec_div.log                153.010 seconds    72.890 seconds        73.350 seconds           148.790 seconds          60.670 seconds              
./cuda/test_float_vec_exp.log                188.430 seconds    92.170 seconds        87.420 seconds           179.780 seconds          72.700 seconds              
./cuda/test_float_vec_log.log                185.530 seconds    88.310 seconds        95.090 seconds           185.270 seconds          72.560 seconds              
./cuda/test_float_vec_mul.log                82.480 seconds     61.520 seconds        58.340 seconds           72.980 seconds           48.300 seconds              
./cuda/test_float_vec_sqrt.log               143.290 seconds    82.050 seconds        85.970 seconds           138.240 seconds          66.750 seconds              
./cuda/test_host_memset.log                  48.870 seconds     47.030 seconds        46.840 seconds           44.540 seconds           37.040 seconds              
./cuda/test_log_softmax.log                  354.990 seconds    134.620 seconds       132.180 seconds          328.980 seconds          107.810 seconds             
./cuda/test_matrix_mul.log                   182.480 seconds    89.290 seconds        88.100 seconds           178.460 seconds          73.990 seconds              
./cuda/test_matrix_mul_shared_mem.log        362.110 seconds    165.640 seconds       162.450 seconds          333.380 seconds          126.450 seconds             
./cuda/test_max_pool2d.log                   52.760 seconds     47.640 seconds        46.670 seconds           46.960 seconds           38.780 seconds              
./cuda/test_multiple_binary_load.log         82.830 seconds     83.320 seconds        83.150 seconds           74.330 seconds           66.930 seconds              
./cuda/test_scalar_print.log                 51.590 seconds     45.930 seconds        47.400 seconds           47.100 seconds           37.100 seconds              
./cuda/test_shared_mem_load_store.log        133.300 seconds    87.180 seconds        84.800 seconds           119.570 seconds          67.170 seconds              
./cuda/test_shared_mem.log                   54.790 seconds     49.950 seconds        46.140 seconds           52.350 seconds           43.800 seconds              
./cuda/test_softmax.log                      220.120 seconds    102.610 seconds       103.440 seconds          206.960 seconds          81.540 seconds              
./cuda/test_stack_load.log                   49.810 seconds     48.090 seconds        44.690 seconds           42.090 seconds           42.140 seconds              
./cuda/test_tile_info.log                    76.770 seconds     49.540 seconds        47.640 seconds           74.680 seconds           39.340 seconds              
./cuda/test_vec_add.log                      76.400 seconds     56.260 seconds        55.260 seconds           64.730 seconds           46.220 seconds              
./cuda/test_vec_add_parallel.log             86.950 seconds     59.010 seconds        58.800 seconds           78.880 seconds           48.600 seconds              
./cuda/test_vec_add_parallel_multi_grid.log  97.040 seconds     66.270 seconds        66.190 seconds           94.340 seconds           52.910 seconds              
./cuda/test_vec_add_serial_multi_grid.log    101.190 seconds    65.550 seconds        69.210 seconds           91.870 seconds           52.690 seconds              
./cuda/test_vec_add_shared_mem.log           75.640 seconds     55.770 seconds        53.340 seconds           69.470 seconds           43.640 seconds              
./library/test_manycore_alignment.log        0.640 seconds      0.520 seconds         0.650 seconds            0.510 seconds            0.550 seconds               
./library/test_manycore_credits.log          6.160 seconds      6.260 seconds         6.150 seconds            5.540 seconds            5.100 seconds               
./library/test_manycore_dmem_read_write.log  0.540 seconds      0.590 seconds         0.600 seconds            0.630 seconds            0.630 seconds               
./library/test_manycore_dram_read_write.log  80.910 seconds     71.050 seconds        70.450 seconds           60.400 seconds           58.730 seconds              
./library/test_manycore_eva_read_write.log   61.840 seconds     59.110 seconds        56.480 seconds           49.110 seconds           46.270 seconds              
./library/test_manycore_init.log             0.650 seconds      0.660 seconds         0.810 seconds            0.610 seconds            0.530 seconds               
./library/test_manycore_packets.log          0.530 seconds      0.500 seconds         0.650 seconds            0.560 seconds            0.440 seconds               
./library/test_manycore_vcache_sequence.log  74.340 seconds     74.880 seconds        72.910 seconds           63.990 seconds           61.220 seconds              
./library/test_printing.log                  0.480 seconds      0.460 seconds         0.550 seconds            0.640 seconds            0.380 seconds               
./library/test_read_mem_scatter_gather.log   0.610 seconds      0.580 seconds         0.710 seconds            0.510 seconds            0.540 seconds               
./library/test_rom.log                   i   0.580 seconds      0.550 seconds         0.490 seconds            0.410 seconds            0.470 seconds               
./library/test_struct_size.log               0.430 seconds      0.690 seconds         0.480 seconds            0.420 seconds            0.400 seconds               
./library/test_vcache_flush.log              53.040 seconds     54.480 seconds        53.580 seconds           46.130 seconds           41.850 seconds              
./library/test_vcache_sequence.log           30.700 seconds     31.810 seconds        28.920 seconds           24.490 seconds           24.010 seconds              
./library/test_vcache_simplified.log         0.710 seconds      0.630 seconds         0.660 seconds            0.620 seconds            0.650 seconds               
./library/test_vcache_stride.log             1.160 seconds      1.090 seconds         1.090 seconds            0.880 seconds            0.960 seconds               
./python/test_python.log                     0.480 seconds      0.490 seconds         0.400 seconds            0.430 seconds            0.450 seconds               
./spmd/test_bsg_dram_loopback_cache.log      1.510 seconds      1.590 seconds         1.570 seconds            1.550 seconds            1.400 seconds               
./spmd/test_bsg_loader_suite.log             15.460 seconds     15.470 seconds        14.360 seconds           14.610 seconds           12.310 seconds              
./spmd/test_bsg_print_stat.log               1.660 seconds      1.640 seconds         1.610 seconds            1.600 seconds            1.600 seconds               
./spmd/test_bsg_scalar_print.log             1.380 seconds      2.360 seconds         1.490 seconds            1.220 seconds            1.250 seconds               
./spmd/test_fib.log                          12.050 seconds     8.970 seconds         9.000 seconds            10.040 seconds           7.530 seconds               
./spmd/test_putchar_stream.log               6.750 seconds      3.590 seconds         4.000 seconds            6.010 seconds            3.100 seconds               
./spmd/test_symbol_to_eva.log                0.490 seconds      0.530 seconds         0.470 seconds            0.590 seconds            0.400 seconds               
```

drichmond:
debug_pp_vpd_trace (slowest) = -debug_pp (a compile flag, necessary for vpd generation), .vpd generated, trace  generated
debug_pp_vpd_no_trace = -debug_pp, .vpd generated, no trace
debug_pp_no_vpd_no_trace = -debug_pp, no .vpd generated, no trace
no_debug_pp_no_vpd_trace = No -debug_pp, no .vpd, trace generated
no_debug_pp_no_vpd_no_trace (fastest) = No debug_pp, no .vpd, no trace

What I'm seeing is that .vpd files slow down simulation, but traces slow it down way more. @vb000 Does this confirm your numbers?

drichmond:
Also thanks for posting these!

drichmond:
Does anyone have any objections to adding suffix .cosim to the cosimulation executables?

taylor-bsg:
- I believe the performance impact of printing the traces is smallish, but
some tests are slowed down excessively because some tests are hyperactive
about print_stat calls and the actual program slows down significantly from
network traffic congestion?

@bornaehsani can confirm?

bornaehsani:
If there is an excessive number of print_stats, yes it will result in a significant slowdown. In the barrier_loop test, where the kernel performs a barrier 64 times in a loop, adding a print_stat to the inner loop (so each tile does 64 print_stats) in a 4x4 tile group, more that quadruples the execution time, and creates terrible network congestion. But, as long as there is a reasonable number of print_stats, for example 2 or 3 print stats in a kernel, then the impact is minimal. 
Also, the results that are shown here, are from the regression directory of the bsg_f1, none of those tests have a print_stat baked on by default, so if the results are from the master branch the slowdown cannot be as a result of the print_stat instructions. 
The trace itself is another issue, if the TRACE is on, it is printed on every cycle regardless of whether there are print_stat instructions in the kernel or not, as to that, I am not sure how much they affect the cosim time as I haven't tried running cosim without the TRACES.

drichmond:
What Borna says is consistent with my understanding. 

Another confounding factor may be that when regression runs, all of the tests within a directory (cuda, spmd, library) write to the same generic trace files. This means that there is between 10-20 open processes writing to the same file. 

I don't know how the OS handles this. It could be falling on its face.

drichmond:
Here's my sense at the moment:

It's unusual to want a trace AND a waveform. The normal case is either a trace, or no trace, or a waveform.

Does anyone disagree?

Therefore, these should be the normal case, and they should be fast and intuitive.

taylor-bsg:
Seems like we should fix this? Generally the best practice for this has
been to have a separate run / build directory for each program that runs,
rather than trying to have unique file names.

M

drichmond:
Yeah, I have a solution. Porting it from baseline (or whatever we decide to call it)

drichmond:
Closing this, superseded by #509

PR #514, opened 2020-03-04
Reuse CUDA init_custom_dimension in global init

drichmond:
This is a PR from a long time ago. I increased code reuse.

It does rely on passing 0,0 to the initializer. If this controversial, let me know. I did comment it though


drichmond:

PR #524, opened 2020-03-18
Add support for machine-specific VCS worklibs 

drichmond:
* Renamed cosim_wrapper.sv to machine_wrapper.sv
  * Renamed the module to use `BSG_MACHINE_NAME macro instead
* Changed VCS Worklib compilation process produce a machine-specific worklib
* Modified synopsys_sim.setup generation process to handle above
* Removed default Makefile.machine.include in the repository root
  * Users now modify path in machine.mk
* Split BSG_MACHINE_PATH from environment.mk into machine.mk
  * Users now modify path in machine.mk
* Updated README

The goal here is to painlessly support switching between different types of machines. I had trouble doing this yesterday with Emily, so undoubtably other users will as well. 

The major change is that we no longer have a "default" Makefile.machine.include in the root of the repository. Instead, BSG_MACHINE_PATH is defined in machine.mk. Everything else is invisible to the user.

Of course, though, the devil is in the details.


drichmond:
This passes regression on all machines

testbenches/dramsim3.mk
```diff
@@ -126,10 +126,10 @@ $(TESTBENCH_PATH)/libdramsim3.so: $(BASEJUMP_STL_DIR)/bsg_test/bsg_dramsim3.cpp
 	$(CXX) $(CXXFLAGS) $(INCLUDES) $^ -Wl,-soname,$(notdir $@) -o $@
 
 endif # ifneq ($(filter $(_DRAMSIM3_MEM_CFGS), $(CL_MANYCORE_MEM_CFG)),)
-endif # ifndef(_BSG_F1_TESTBENCHES_DRAMSIM3_MK)
```
mrutt92: is this undo-ing another PR?

drichmond: Yes/no. I was too aggressive in #526 and I moved it outside of the ifdef. This just fixes that.

PR #525, opened 2020-03-18
adds dma/backdoor for infinite memory configuration

mrutt92:
adds dma/backdoor for infinite memory configuration


testbenches/infmem.mk
```diff
@@ -0,0 +1,81 @@
...
+
+# CL_DIR: The path to the root of the BSG F1 Repository
+ifndef CL_DIR
+$(error $(shell echo -e "$(RED)BSG MAKE ERROR: CL_DIR is not defined$(NC)"))
+endif
+
+# TESTBENCH_PATH: The path to the testbenches folder in BSG F1
+ifndef TESTBENCH_PATH
+$(error $(shell echo -e "$(RED)BSG MAKE ERROR: TESTBENCH_PATH is not defined$(NC)"))
+endif
```
drichmond: I think you only need to check testbench path

testbenches/libdmamem.mk
```diff
@@ -0,0 +1,85 @@
...
+# LIBRARIES_PATH: The path to the regression folder in BSG F1
+ifndef LIBRARIES_PATH
+$(error $(shell echo -e "$(RED)BSG MAKE ERROR: LIBRARIES_PATH is not defined$(NC)"))
+endif
+
+# PROJECT: The project name, used to as the work directory of the hardware
+# library during analysis
+ifndef PROJECT
+$(error $(shell echo -e "$(RED)BSG MAKE ERROR: PROJECT is not defined$(NC)"))
+endif
```
drichmond: You should replace these with a check for basejump

mrutt92:
you already have to do that...

drichmond:
You do? 

We should fix that then. I didn't realize that was an implication of an earlier PR.

drichmond:
As we discussed this morning, and as solved in #524, it's annoying to run `make clean` when switching between machines. The runtime library should be smart enough to know when certain features aren't available (granted, in cosimulation only)

It seems like a solution would be to add bits in the rom for certain DMA features. Is my sense correct?

mrutt92:
It seems reasonable - although proper makefile dependency listing would defect solve the problem we discussed this morning.

hardware/hardware.mk
```diff
@@ -144,6 +144,23 @@ $(HARDWARE_PATH)/%.v: $(HARDWARE_PATH)/%.rom
...
+CL_MANYCORE_MEMSYS_FEATURE_DMA   := 1
+CL_MANYCORE_MEMSYS_FEATURE_CACHE := 1
+endif
+
+# do we use infinite memory?
+ifneq ($(filter infinite, $(subst _, ,$(CL_MANYCORE_MEM_CFG))),)
+CL_MANYCORE_MEMSYS_FEATURE_DMA   := 1
+CL_MANYCORE_MEMSYS_FEATURE_CACHE := 0
+endif
+
```
drichmond: Thanks! 

I think there's still a discussion to be had about where we put memory-specific definitions but for now this is great!

testbenches/cosimulation.mk
```diff
@@ -191,20 +191,30 @@ $(REGRESSION_TESTS:%=.%/vanilla_stats.csv): .%/vanilla_stats.csv : %.rule
...
+# set the the name for MEMSYS_STATS
+ifneq ($(filter e_vcache_%, $(CL_MANYCORE_MEM_CFG)),)
+MEMSYS_STATS := vcache_stats
+endif
+
+ifneq ($(filter e_infinite_mem, $(CL_MANYCORE_MEM_CFG)),)
+MEMSYS_STATS := infinite_mem_stats
+endif
+
+%.vpd: .%/vcdplus.vpd %.vanilla_stats.csv %.$(MEMSYS_STATS).csv
```
drichmond: Thanks for doing this. Part of me wonders though, should we just unify on some generic name like `memory_system_stats`?

testbenches/dramsim3.mk
```diff
@@ -108,7 +111,6 @@ $(TESTBENCH_PATH)/libdramsim3.so: CXXFLAGS += -std=c++11 -D_GNU_SOURCE -Wall -fP
 $(TESTBENCH_PATH)/libdramsim3.so: CXXFLAGS += -I$(BASEJUMP_STL_DIR)/imports/DRAMSim3/src
 $(TESTBENCH_PATH)/libdramsim3.so: CXXFLAGS += -I$(BASEJUMP_STL_DIR)/imports/DRAMSim3/ext/headers
 $(TESTBENCH_PATH)/libdramsim3.so: CXXFLAGS += -I$(BASEJUMP_STL_DIR)/imports/DRAMSim3/ext/fmt/include
-$(TESTBENCH_PATH)/libdramsim3.so: CXXFLAGS += -I$(BASEJUMP_STL_DIR)/bsg_test
```
drichmond: Are there no longer includes in bsg_test (now called bsg_mem?)

testbenches/cosimulation.mk
mrutt92: that might imply that they have the same format - which I'm not sure is true.

drichmond: Good point

testbenches/dramsim3.mk
mrutt92: the header file was moved to bsg_mem and it's now part of a separate library called libdmamem.

testbenches/cosimulation.mk
drichmond: @tommydcjung or @bornaehsani Do the different memory system files have the same format? Presumably no?

mrutt92:
merge?

PR #527, opened 2020-03-18
Removed all references to ramulator

drichmond:


testbenches/simlibs.mk
```diff
@@ -97,7 +97,6 @@ VSOURCES += $(BASEJUMP_STL_DIR)/bsg_mem/bsg_nonsynth_mem_1rw_sync_assoc.v
 
 VSOURCES += $(BASEJUMP_STL_DIR)/bsg_test/bsg_nonsynth_clock_gen.v
 VSOURCES += $(BASEJUMP_STL_DIR)/bsg_test/bsg_nonsynth_reset_gen.v
-VSOURCES += $(BASEJUMP_STL_DIR)/bsg_test/bsg_nonsynth_ramulator_hbm.v
```
tommydcjung: is the ramulator going to be removed from the cl_manycore?

drichmond: I suppose it should, and I didn't do that.

drichmond: Fixed

drichmond:
This is rebased and ready for review. @mrutt92 How do I run with DRAM Sim3? Just copy the machine file from machines?

drichmond:
I changed the default Makefile.machine.include, currently getting errors like this:

```Error-[SV-EEM-SRE] Scope resolution error
/mnt/users/spin0/no_backup/drichmond/Research/repositories/git/bsg_bladerunner/bsg_replicant/hardware/cl_manycore.sv, 900
lv3_dramsim3, "bsg_dramsim3_e_vcache_blocking_test_dramsim3_hbm2_4gb_x128_pkg::num_columns_p"
  Target for scope resolution operator does not exist. Token
  'bsg_dramsim3_e_vcache_blocking_test_dramsim3_hbm2_4gb_x128_pkg' is not a
  class/package.  Originating module 'cl_manycore'.
  Check that class or package exists with referred token as the name.


Error-[SV-EEM-SRE] Scope resolution error
/mnt/users/spin0/no_backup/drichmond/Research/repositories/git/bsg_bladerunner/bsg_replicant/hardware/cl_manycore.sv, 901
lv3_dramsim3, "bsg_dramsim3_e_vcache_blocking_test_dramsim3_hbm2_4gb_x128_pkg::address_mapping_p"
  Target for scope resolution operator does not exist. Token
  'bsg_dramsim3_e_vcache_blocking_test_dramsim3_hbm2_4gb_x128_pkg' is not a
  class/package.  Originating module 'cl_manycore'.
  Check that class or package exists with referred token as the name.


Error-[SV-EEM-SRE] Scope resolution error
/mnt/users/spin0/no_backup/drichmond/Research/repositories/git/bsg_bladerunner/bsg_replicant/hardware/cl_manycore.sv, 902
lv3_dramsim3, "bsg_dramsim3_e_vcache_blocking_test_dramsim3_hbm2_4gb_x128_pkg::size_in_bits_p"
  Target for scope resolution operator does not exist. Token
  'bsg_dramsim3_e_vcache_blocking_test_dramsim3_hbm2_4gb_x128_pkg' is not a
  class/package.  Originating module 'cl_manycore'.
  Check that class or package exists with referred token as the name.
```

Makefile.machine.include
```diff
@@ -56,8 +56,8 @@ CL_MANYCORE_HOST_COORD_Y             := 1
 CL_MANYCORE_DIM_X                    := $(BSG_MACHINE_GLOBAL_X)
 CL_MANYCORE_DIM_Y                    := $(shell expr $(BSG_MACHINE_GLOBAL_Y) - 1)
 
-CL_MANYCORE_MEM_CFG                  :=e_vcache_blocking_axi4_f1_model
-CL_MANYCORE_DRAM_CHANNELS            :=1
+CL_MANYCORE_MEM_CFG                  := e_vcache_blocking_test_dramsim3_hbm2_4gb_x128
```
tommydcjung: can we drop _4gb_x128, since we will not be using other configs.

mrutt92: I think we should leave it. We might use another config in the future.

hardware/cl_manycore.sv
```diff
@@ -897,37 +890,8 @@ module cl_manycore
...
-       );
-
-  end // block: lv3_ramulator_hbm
-  else if (mem_cfg_p == e_vcache_blocking_dramsim3_hbm2_4gb_x128 ||
-           mem_cfg_p == e_vcache_non_blocking_dramsim3_hbm2_4gb_x128) begin: lv3_dramsim3
+  else if (mem_cfg_p == e_vcache_blocking_test_dramsim3_hbm2_4gb_x128 ||
+           mem_cfg_p == e_vcache_non_blocking_test_dramsim3_hbm2_4gb_x128) begin: lv3_dramsim3
 `ifdef USING_DRAMSIM3
     bsg_nonsynth_dramsim3
       #(.channel_addr_width_p(`DRAMSIM3_MEM_PKG::channel_addr_width_p)
```
mrutt92: Make sure that DRAMSIM3_MEM_PKG is set correctly in `test benches/dramsim3.mk`

drichmond: +define+DRAMSIM3_MEM_PKG=bsg_dramsim3_e_vcache_blocking_test_dramsim3_hbm2_4gb_x128_pkg

mrutt92: needs to be one of the packages in this file https://github.com/bespoke-silicon-group/basejump_stl/blob/master/bsg_test/bsg_dramsim3_pkg.v

drichmond: Thanks Max! That led me to the issue.

drichmond:
This passes regression on e_vcache_blocking_test_dramsim3_hbm2_4gb_x128

hardware/cl_manycore.sv
drichmond: I think Tommy's question is a good one to consider. Is there another way we can transform one enum into the other that doesn't use the makefiles? I worry a little about name instability

drichmond: But I don't think it should be addressed in this PR. If we can summarize it or come up with a solution we should file an issue

tommydcjung:
```
if (enum == e_something)
  `define DRAMSIM3_MEM_PKG e_a
else if (enum == e_something_else)
  `define DRAMSIM3_MEM_PKG e_b
```
does this work?

drichmond: It would, but something more generic would be appreciated. If we do it this way, every single time that we add a new DRAMSim3 memory configuration we have to modify the verilog. 

Ideally we don't have to do that, and that's the nice part of the current Makefile solution

drichmond: The lines that we're discussing:


https://github.com/bespoke-silicon-group/bsg_replicant/blob/b0b04eeceead35740a14191e48e91bc65d484034/testbenches/dramsim3.mk#L81-L89

tommydcjung: The downside of that approach is that every time mem_cfg_p changes this make also needs to change? also if there is other mem_cfg_p that starts with something other than "e_vcache_blocking_" or "e_vcache_non_blocking_" you have to modify the makefile too.

drichmond: Yes to the second, but I'm not sure about the first. Doesn't this handle mem_cfg_p changes?

mrutt92: this occurred because he changed the base name of the config for dramsim3.

drichmond:
Passes regression on both HBM Machines

hardware/cl_manycore.sv
mrutt92: also Tommy I don't think the approach you posted would work? you would need to use `ifdef for conditional preprocessor setting I think.

tommydcjung: My point is that mem_cfg_p already has all the information what the memory configuration should be. We don't need another macro.

drichmond: I think we do? we use it to configure the memory system of replicant

tommydcjung:
```
if (enum == e_4gb) begin: hbm_4gb
  localparam density_lp = 4gb_pkg::density;
end
else if (enum == e_8gb) begin: hbm_8gb
  localparam density_lp = 8gb_pkg::density;
end
```

mrutt92: does density_lp lose it's scope?

drichmond: OK, I'm going to merge this but I think there's some serious value in this discussion because it may allow us to be more extensible. I see value, but I don't understand the (engineering) cost. We should move this discussion to an issue.

drichmond:
@mrutt92, you have made me laugh today. While it was not part of your "mission statement" I commend you.

hardware/cl_manycore.sv
tommydcjung:
```
if (mem_cfg_p == e_4gb) begin
  import 4gb_pkg::*;
end
else if (mem_cfg_p == e_8gb) begin
  import 8gb_pkg::*;
end

bsg_module #(
 .density_p(hbm2_density_p)
) m0 (
);
```
this would require change parameter names in those packages to something that wouldn't cause name conflict.

PR #528, opened 2020-03-23
Fix Python Origin issues

drichmond:
* Mirrored changed parameter name in cl_manycore.sv


drichmond:
Should be merged with https://github.com/bespoke-silicon-group/bsg_manycore/pull/266

PR #529, opened 2020-03-26
One cache one bank

mrutt92:
* caches are mapped to separate banks
* adds memory system information to the ROM
* memory system configuration is modularized in the makefiles


mrutt92:
needs https://github.com/bespoke-silicon-group/basejump_stl/pull/262

drichmond:
I like this, but going to spend some more time reviewing it in detail if you don't mind

hardware/cl_manycore.sv
```diff
@@ -893,6 +893,42 @@ module cl_manycore
   else if (mem_cfg_p == e_vcache_blocking_test_dramsim3_hbm2_4gb_x128 ||
            mem_cfg_p == e_vcache_non_blocking_test_dramsim3_hbm2_4gb_x128) begin: lv3_dramsim3
 `ifdef USING_DRAMSIM3
```
drichmond: Not part of this PR, but why do we need to have USING_DRAMSIM3 if we only instantiate it when it is specified by mem_cfg_p?

mrutt92:
maybe review by voice to streamline this? it's needed for the memory system configurations outlined in the HBTRM.

drichmond:
> maybe review by voice to streamline this? it's needed for the memory system configurations outlined in the HBTRM.

Yeah, let's do it after the meeting

hardware/memsys.mk
```diff
@@ -0,0 +1,57 @@
...
+# memory hierarchy bitidx
+CL_MANYCORE_MEMSYS_DRAM_RO_BITIDX := 0
+CL_MANYCORE_MEMSYS_DRAM_BG_BITIDX := 0
+CL_MANYCORE_MEMSYS_DRAM_BA_BITIDX := 0
+CL_MANYCORE_MEMSYS_DRAM_CO_BITIDX := 0
+CL_MANYCORE_MEMSYS_DRAM_BYTE_OFF_BITIDX := 0
+
+include $(HARDWARE_PATH)/memsys_infinite_mem.mk
+include $(HARDWARE_PATH)/memsys_dramsim3.mk
+include $(HARDWARE_PATH)/memsys_axi4_f1.mk
```
drichmond: Discussion question for later - perhaps we should put the memory system fragments in a separate folder.

drichmond:
Should lines like these be part of the memory system definition, or part of simlibs?

https://github.com/bespoke-silicon-group/bsg_replicant/blob/9d3baeb8b3f4b22a6a1fa222575a60ed19ee7ada/testbenches/simlibs.mk#L153-L160

drichmond:
You'll also need to rebase on master

hardware/hardware.mk
```diff
@@ -145,25 +151,31 @@ $(HARDWARE_PATH)/%.v: $(HARDWARE_PATH)/%.rom
...
-# do we use infinite memory?
-ifneq ($(filter infinite, $(subst _, ,$(CL_MANYCORE_MEM_CFG))),)
-CL_MANYCORE_MEMSYS_FEATURE_DMA   := 1
-CL_MANYCORE_MEMSYS_FEATURE_CACHE := 0
-endif
+include $(HARDWARE_PATH)/memsys.mk
+
+.PHONY: charvtobin-debug
+charvtobin-debug:
+	@echo $(CL_MANYCORE_MEMSYS_ID) = $(call charv2bin, $(CL_MANYCORE_MEMSYS_ID))
```
drichmond: This is probably useful to leave in, but just to make sure, this was your intent?

libraries/bsg_manycore_config.h
```diff
@@ -104,8 +107,7 @@ extern "C" {
                 uint32_t io_endpoint_max_out_credits;
                 uint32_t dram_channels;
                 uint32_t dram_bank_size_words;
```
drichmond: Should these be part of the memory system too?

hardware/cl_manycore.sv
mrutt92: It's from a few PRs back and I don't remember exactly. Could have been the wrong solution to some problem I was running into.

drichmond: We don't necessarily need to solve it now, but I would like to run it to ground, so to speak. You can file an issue or we can discuss it later

hardware/hardware.mk
mrutt92: It was not - I'm going to remove it.

libraries/bsg_manycore_config.h
mrutt92: Seems reasonable.

mrutt92:
@drichmond want to take a final look?

drichmond:
Does it pass?

PR #533, opened 2020-03-27
Creates Machine Configurations outlined in Arch Description of HB-TRM

mrutt92:
* FAST_N_FAKE
* TIMING_V0_8_4
* TIMING_V0_16_8
* TIMING_V0_32_16
* TIMING_V0_64_32
* Extends test_rom to sanity-check DRAM and bank size 
* Fixes bug introduced in #529 - RISCV linker script needs BSG_MACHINE_DRAM_BANK_SIZE defined in Makefile.machine.include


machines/4x4_fast_n_fake
```diff
@@ -0,0 +1 @@
+4x4_infinite_memory
```
drichmond: I think we should just have one version, instead of a symlink.

mrutt92:
The linker error addressed in this PR was caused by these two lines expecting that BSG_MACHINE_DRAM_BANK_SIZE is defined in `$(BSG_MACHINE_PATH)/Makefile.machine`

https://github.com/bespoke-silicon-group/bsg_manycore/blob/1175de21508b53d436b2676f1a230048bbe5acae/software/mk/Makefile.builddefs#L29

https://github.com/bespoke-silicon-group/bsg_manycore/blob/1175de21508b53d436b2676f1a230048bbe5acae/software/mk/Makefile.builddefs#L153

mrutt92:
I think this is good to merge

drichmond:
For posterity - we decided that it was not a good idea to have symlinks because it may cause `multiverse` regression to run duplicates

PR #537, opened 2020-03-30
Refactor/Remove AWS File List

drichmond:
I'm working to understand the TB hierarchy of the AWS code, so that someone can use verilator, and we have a better understanding of the clock generator (from AWS)

This PR splits up the AWS files. Some VINCLUDES/VSOURCES were moved to the memsys_f1_axi.mk, some were moved into simlibs, and the remaining AWS-specific ones were put in a new aws.mk fragment. 

The two encrypted files/modules that block verilator*:
- aws-fpga/hdk/common/shell_stable/design/interfaces/unused_ddr_a_b_d_template.inc: This includes the sh_ddr module to tie off the unused DDR inputs. sh_ddr is an encrypted module
- hardware/cl_manycore.sv: Contains axi_clock_converter, which is an encrypted IP

It might be possible to tie of the unused DDR in a different way, without using unused_ddr_a_b_d_template.inc, and the AXI clock converter isn't strictly necessary.

\* As far as I know. There could be more.


drichmond:
Outdated/stale

PR #538, opened 2020-03-31
Added support for running Manycore at HBM frequency

drichmond:
As title says, fixed the manycore clock frequency using the USING_DRAMSIM3 macro

No idea if this is the right way to do it, yet. Will test it out

### DIFF
--- a/hardware/cl_manycore.sv
+++ b/hardware/cl_manycore.sv
@@ -65,8 +65,6 @@ module cl_manycore
 `include "unused_sh_bar1_template.inc"
 `include "unused_apppf_irq_template.inc"
 
-   localparam lc_clk_main_a0_p = 8000; // 8000 is 125 MHz
-
    //-------------------------------------------------
    // Wires
    //-------------------------------------------------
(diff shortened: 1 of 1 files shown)


hardware/cl_manycore.sv
```diff
@@ -754,11 +765,7 @@ module cl_manycore
 
     //500MHz
     bsg_nonsynth_clock_gen
-`ifdef USING_DRAMSIM3
       #(.cycle_time_p(`DRAMSIM3_MEM_PKG::tck_ps))
-`else
-      #(.cycle_time_p(2000))
-`endif
```
drichmond: Also removed this ramulator hold-over

drichmond:
Still working on testing this. Given that it's simple though, open to suggestions

drichmond:
Running regression but I verified the clock frequency: 
image

drichmond:
Passes regression on timing_v0_16_8

drichmond:
@tommydcjung, your thoughts?

tommydcjung:
It looks okay, but I still think that the clock-mux thing should be gone.

drichmond:
probably, but not part of this PR.

PR #548, opened 2020-04-03
Regression now produces runtime measurements

drichmond:
This is probably long overdue, but running regression from testbenches now produces a runtime.log file that records the execution time (as reported by VCS) of each testbench.

That runtime.log file is copied into BSG_MACHINE_PATH, so that running regression on multiple machines produces runtime.log files for each machine. 

Anything else that people want in this? @proftaylor 

For example, I might be able to report manycore core clock cycles.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,14 +66,17 @@ regression:
 cosim: 
 	$(MAKE) -C testbenches regression
 
-__BSG_MACHINES := $(wildcard machines/*)
-__BSG_MACHINES := $(filter-out machines/README.md,$(__BSG_MACHINES))
-__BSG_MACHINES := $(filter-out machines/timing_v0_32_16,$(__BSG_MACHINES))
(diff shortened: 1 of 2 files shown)


drichmond:
For example here's a graph where I tried to conclude if clock switching on IO is still effective: 
image

On the left axis, is the ratio of runtimes for 41 Cuda Regression tests, computed as `Switching Runtime/Non-Switching Runtime` (A ratio of 1 indicates that the runtimes were equal, greater than 1 indicates that the Switching runtime was longer than the non-switching runtime, and less than 1 indicates that the switching runtime was shorter than the switching runtime). The Y axis are the five different machines I collected results from. 

I generated this graph because I was hoping to demonstrate my hypothesis that Clock-Switching is still beneficial. I believe this graph suggests that clock switching is beneficial on the infinite memory designs, but more investigation is required. 


(I don't like that the Y axis on this graph isn't linear, but I couldn't think of a better way to display this data)

proftaylor:
Neat!

PR #558, opened 2020-04-20
Python cmd args

vb000:
- Added support for arbitrary python command line arguments in test_loader.


vb000:
Yes Dustin, test_python in the regression suite can be done this way... if we are okay with breaking the `+c_args="<test_dir> <test_name>"` invariant.

drichmond:
OK. If it only applies to test_python I'm not concerned. 

Looks good to me, do you want me to merge it?

vb000:
Sure.

vb000:
> OK. If it only applies to test_python I'm not concerned.
> 
> Looks good to me, do you want me to merge it?

Let's make this change in test_python then?

drichmond:
You can if you want. I added test_python mostly to show how to integrate python. Nobody has added additional functionality, so I'm not concerned with the way it is right now.

vb000:
Okay, just to have the flexibility in case we need it, I changed it in #562.

PR #559, opened 2020-04-20
Minor refactoring of the Endpoint-To-FIFO module

drichmond:
Moving to Verilator, this module was slightly problematic. Several parameters had unnecessary global scope, and were included in a parameter file that was outside of the context of the endpoint-to-fifo module.

In addition, it seems like we had duplicate stat-print snoop logic. I removed that, as well.

* Removed some global parameters, that could be computed from local parameters (!!!)
* Removed Endpoint-To-FIFO dependence on AXI-L parameters
* Removed duplicate statistics snooping logic

### DIFF
--- a/hardware/bsg_manycore_endpoint_to_fifos.v
+++ b/hardware/bsg_manycore_endpoint_to_fifos.v
@@ -33,20 +33,21 @@
 * Or cast the manycore packet to the rx_fifo data stream
 */
 
-
 module bsg_manycore_endpoint_to_fifos
-  import cl_manycore_pkg::*;
   import bsg_manycore_pkg::*;
-  import bsg_manycore_link_to_axil_pkg::*;
-  import bsg_manycore_addr_pkg::bsg_print_stat_epa_gp;
 #(
   parameter fifo_width_p = "inv"
   // these are endpoint parameters
   , parameter x_cord_width_p = "inv"
+  , localparam x_cord_width_pad_lp = `BSG_CDIV(x_cord_width_p,8)*8
   , parameter y_cord_width_p = "inv"
+  , localparam y_cord_width_pad_lp = `BSG_CDIV(y_cord_width_p,8)*8
   , parameter addr_width_p = "inv"
+  , localparam addr_width_pad_lp = `BSG_CDIV(addr_width_p,8)*8
   , parameter data_width_p = "inv"
+  , parameter data_width_pad_lp = `BSG_CDIV(data_width_p,8)*8
   , parameter max_out_credits_p = "inv"
+  , parameter ep_fifo_els_p = "inv"
   , parameter link_sif_width_lp = `bsg_manycore_link_sif_width(addr_width_p,data_width_p,x_cord_width_p,y_cord_width_p)
 ) (
   input                                      clk_i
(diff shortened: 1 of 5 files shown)


drichmond:
Passes fast_n_fake cosimulation regression

hardware/bsg_manycore_endpoint_to_fifos.v
```diff
@@ -33,20 +33,21 @@
 * Or cast the manycore packet to the rx_fifo data stream
 */
 
-
 module bsg_manycore_endpoint_to_fifos
-  import cl_manycore_pkg::*;
   import bsg_manycore_pkg::*;
-  import bsg_manycore_link_to_axil_pkg::*;
-  import bsg_manycore_addr_pkg::bsg_print_stat_epa_gp;
```
drichmond, line 41: This was only needed for stat snooping, which should not be part of this module. It is encapsulated in https://github.com/bespoke-silicon-group/bsg_replicant/blob/master/hardware/bsg_print_stat_snoop.v

(I believe)

hardware/bsg_manycore_endpoint_to_fifos.v
```diff
@@ -33,20 +33,21 @@
...
+  , localparam x_cord_width_pad_lp = `BSG_CDIV(x_cord_width_p,8)*8
   , parameter y_cord_width_p = "inv"
+  , localparam y_cord_width_pad_lp = `BSG_CDIV(y_cord_width_p,8)*8
   , parameter addr_width_p = "inv"
+  , localparam addr_width_pad_lp = `BSG_CDIV(addr_width_p,8)*8
   , parameter data_width_p = "inv"
+  , parameter data_width_pad_lp = `BSG_CDIV(data_width_p,8)*8
   , parameter max_out_credits_p = "inv"
+  , parameter ep_fifo_els_p = "inv"
   , parameter link_sif_width_lp = `bsg_manycore_link_sif_width(addr_width_p,data_width_p,x_cord_width_p,y_cord_width_p)
```
drichmond: These used to be _gp but that seems wildly unnecessary, since they're only used locally. It also removes the dependence on bsg_manycore_link_to_axil_pkg.

hardware/bsg_manycore_endpoint_to_fifos.v
```diff
@@ -77,13 +78,9 @@ module bsg_manycore_endpoint_to_fifos
 
   ,output [`BSG_WIDTH(max_out_credits_p)-1:0] out_credits_o
 
-
-  // print stat
-  ,output                                     print_stat_v_o
-  ,output [                 data_width_p-1:0] print_stat_tag_o
```
drichmond, line 83: Duplicant functionality in bsg_manycore_stat_snoop

hardware/bsg_manycore_endpoint_to_fifos.v
leonardxiang: I agree

hardware/bsg_manycore_endpoint_to_fifos_pkg.v
```diff
@@ -0,0 +1,48 @@
...
+    logic [y_cord_width_mp-1:0] y_cord    ; \
+    logic [x_cord_width_mp-1:0] x_cord    ; \
+  } bsg_mcl_request_s
+
+
+package bsg_manycore_endpoint_to_fifos_pkg;
+
+endpackage : bsg_manycore_endpoint_to_fifos_pkg
+
+`endif // BSG_MANYCORE_ENDPOINT_TO_FIFOS_PKG_V
```
leonardxiang, line 48: Since this part is decoupled, I think the rest of _bsg_manycore_link_to_axil_pkg.v_ should be renamed to axil_to_fifos_pkg.v & axil_to_mem_pkg.v.

bornaehsani:
Cosimulation regression passes, all profiling graphs, stats and pc histograms are generated!

hardware/bsg_manycore_endpoint_to_fifos_pkg.v
drichmond: This is a good point

drichmond:
This passes regression on 4x4_fast_n_fake, and timing_v0_8_4

drichmond:
Happily

drichmond:
Passes timing_v0_16_8

PR #560, opened 2020-04-21
adds a type safe cuda memcpy

mrutt92:
Adds type safe alternatives to `hb_mc_device_memcpy`.


drichmond:
I assume these have been tested somewhere?

mrutt92:
tested in cosim with 4x4 fast_n_fake

bornaehsani:
I'm a bit late to the party, but this looks good! Thanks Max!

PR #561, opened 2020-04-21
malloc uses the cache size as the alignment instead of hardcoded 32

mrutt92:
* The memory allocator now aligns based on the victim cache's block size.


mrutt92:
brain fart - I meant malloc, not memcpy.

drichmond:
I know, just giving you a hard time. Thanks for doing this

PR #565, opened 2020-04-22
victim caches now have 32 word cache lines with 16 sets each

mrutt92:
HBM benefits greatly from 32 word cache lines.


dpetrisko:
Interesting, do you think it’s the prefetching effect?

taylor-bsg:
@dpetrisko The basic HBM transaction is 8 (32-bit) words, and you can read
every cycle in a pipelined fashion. However there is a limit of opening 4
new pages per channel per 30 cycles (so-called tFAW limit).  So that means,
if you do random access, you are limited by tFAW to 4 random accesses per
30 cycles. However tFAW does not limit your ability to read data off of
pages you have opened. So essentially, the larger block size (=3 addition
reads) comes for free in the random access case. And it helps you greatly
in sequential access, because it means you do not need to have as many
outstanding cache misses in order to saturate the HBM bandwidth. The only
case where this is suboptimal is if you are doing strided accesses within a
page that are not that close, but it is not clear that that is very common.

PR #566, opened 2020-04-24
Updated Cosimulation Makefile rules to generate vcache stall graph and stats parser

bornaehsani:
Added a <test_name>.vcache_operation_trace.csv Make rule to generate vcache operation trace
Added a <test_name>.operation_trace  Make rule that creates both vanilla_operation_trace.csv and vcache_operation_trace.csv.

Cosim regression passed, files are generated.


bornaehsani:
> Does it work? Tried it?

Oh yeah, forgot to mention it :D Cosim regression passes, and the files are generated as expected!

bornaehsani:
Howdy @drichmond ! I don't have merge access, can you please merge this in if you think it's good to go? Thanks!

PR #568, opened 2020-04-28
Simplified main/cosim main calling

drichmond:
* moved cosim_main to header file (which seems REALLY strange)
* Moved svScope VCS code into bsg_manycore_lib with macro guards

Passes regression on 4x4_fast_n_fake (except for test_symbol_to_eva)

This makes cosim_main VCS specific, not cosim specific (because Verilator uses main)

### DIFF
--- a/libraries/bsg_manycore.cpp
+++ b/libraries/bsg_manycore.cpp
@@ -38,6 +38,7 @@
 #include <fpga_pci.h>
 #include <fpga_mgmt.h>
 #else
+#include <svdpi.h>
 #include <fpga_pci_sv.h>
 #include <utils/sh_dpi_tasks.h>
 #include <bsg_mem_dma.hpp>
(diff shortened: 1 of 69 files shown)


libraries/bsg_manycore.cpp
```diff
@@ -536,6 +561,7 @@ int  hb_mc_manycore_init(hb_mc_manycore_t *mc, const char *name, hb_mc_manycore_
         r = err;
         hb_mc_manycore_cleanup_fifos(mc);
         hb_mc_manycore_cleanup_mmio(mc);
+        hb_mc_manycore_cleanup_dpi(mc);
```
mrutt92, line 564: what happens if COSIM is not defined?

drichmond: Oops, missed that

drichmond: Fixed

regression/library/test_printing.c
```diff
@@ -47,6 +47,15 @@
 #include "test_printing.h"
 static int test_printing(void)
 {
+        int rc;
```
mrutt92, line 50: why is this needed now?

libraries/bsg_manycore.cpp
```diff
@@ -254,6 +258,21 @@ static int hb_mc_manycore_rx_fifo_drain(hb_mc_manycore_t *mc, hb_mc_fifo_rx_t ty
...
+{
+        svScope scope;
+        scope = svGetScopeFromName("tb");
+        svSetScope(scope);
+        return HB_MC_SUCCESS;
+}
+
+static void hb_mc_manycore_cleanup_dpi(hb_mc_manycore_t *mc)
+{
+        return;
```
vb000, line 272: What's the job of the function? Are we going to do something in this later?

drichmond: Yep. For an incoming feature :p 

(All of my current PRs are for a fancy new feature)

regression/library/test_printing.c
drichmond: Fixed

drichmond:
This passes regression with 4x4_fast_n_fake, and timing_v0_8_4

@mrutt92, sufficient for you?

PR #571, opened 2020-04-29
Removed unused files

drichmond:
These files were removed in manycore, so I'm doing the corresponding change here


drichmond:
Passes 4x4_fast_n_fake (which is the only PR this MIGHT affect)

PR #577, opened 2020-05-19
Moved BSG Manycore-specific files to BSG Manycore

drichmond:
I think these files would be better in BSG Manycore. 

bsg_print_stat_snoop.v: Only used in simulation, so I've moved it OUT of the hardware directory (which contains synthesizable hardware) and into bsg_manycore/testbenches/common/v

bsg_manycore_endpoint_to_fifos.v: This isn't F1 specific, and it's actually quite useful. I think it would be a good idea to have this in BSG Manycore in the long run.

The long-term benefit of this is that the hardware directory only contains files that are relevant to FPGA compilation (OK, this isn't strictly true yet, but we'll get there)


drichmond:
Passes regression on fast_n_fake

PR #578, opened 2020-05-20
CUDA Malloc Bugfix in API

bornaehsani:
1. API now resets the number of tile groups to zero in the host, after every hb_mc_tile_groups_execute call.
2. API allocates a space in device DRAM to store kernel arguments before tile group launch, it now frees that location after tile group is executed.
3. Adds a memory_leak regression test to make sure CUDA Lite API does not have memory leak on launching tile groups.

**merge with PR # 317 in bsg_manycore.**

### DIFF
--- a/libraries/bsg_manycore_cuda.cpp
+++ b/libraries/bsg_manycore_cuda.cpp
@@ -642,6 +642,7 @@ static int hb_mc_tile_group_enqueue (hb_mc_device_t* device,
         tg->id = tg_id;
         tg->grid_id = grid_id;
         tg->grid_dim = grid_dim;
+        tg->argc = argc;
         tg->status = HB_MC_TILE_GROUP_STATUS_INITIALIZED;
 
         tg->map = (hb_mc_eva_map_t *) malloc (sizeof(hb_mc_eva_map_t)); 
@@ -748,6 +749,10 @@ static int hb_mc_tile_group_launch (hb_mc_device_t *device,
                 return HB_MC_NOMEM;
         }
 
+        // store the address of argv in the host, to free the 
+        // memory location in the device DRAM after tile group is executed
+        tg->argv_eva = args_eva;
+
         // transfer the arguments to dram
         error = hb_mc_device_memcpy(    device, reinterpret_cast<void *>(args_eva),
                                         (void *) &(tg->kernel->argv[0]),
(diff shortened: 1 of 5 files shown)


drichmond:
Passes regression?

bornaehsani:
My bad! Keep forgetting to mention regression passes :D Yes cosim regression passed!

drichmond:
On more than one machine ?

bornaehsani:
Yes! Passes on timing machines, 4x4_fast_n_fake and 4x4_vcache_blocking_f1_model machines.

mrutt92:
is there a test that exercises this? i.e. one that fails before the bug fix and passes after?

bornaehsani:
Well this bug doesn't cause any tests to fail, it just gradually slows down execution on the chip. In simulation I imagine it would take a very long time to manifest, I don't know if it's a suitable regression test if it takes hours to complete.

dpetrisko:
Not sure of the details here, but could you malloc a huge amount (1/2 of device memory), run some null kernel, then malloc again and see if you run out of device memory?

mrutt92:
+1 to Dan's suggestion.

mrutt92:
Actually that might not work. This bug specifically applies to memory allocated to kernel arguments.

However, you could run one kernel multiple times, allocating a buffer of the same size and freeing it between calls.

You can inspect the address of that allocated buffer - it should always be the same if everything is cleaned up properly.

drichmond:
I think we should have a regression test for this, if we can. What happens if we just have an empty kernel, and keep re-running it with new buffers?

bornaehsani:
OK I finally added a regression test as Max suggested which I think implicitly makes sure this problem doesn't happen again. It allocates large chunks of memory and launches a tile group with them as input arguments, and then frees them. It repeats this process 32 times, and checks to make sure the address of the allocated pieces remains the same every time (i.e. all memory pieces are freed each time). 
Passes regression on multiple machines. Do you guys think this prevents the issue from popping up again?

drichmond:
👍

mrutt92:
It should be fine. Does the test fail without your fix? If you don't know, a quick way to check might be to checkout master and cherry pick from this branch. Then run the test.

bornaehsani:
Great idea Max! The test passed without the fix in CUDA-Lite API, which uncovered a bug in the memory leak regression test. I fixed that, now it fails without the fix and passes with it!

mrutt92:
Well done! LGTM.

PR #579, opened 2020-05-20
Split out machine-specific functionality into new file

drichmond:
I've introduced the idea of platforms. Platforms are the host medium for running a manycore, be it FPGA, VCS, or Verilator. This PR splits `bsg_manycore.cpp` into two files: 

`bsg_manycore.cpp`: Manycore API - data copy in/out, packet tx/rx
`bsg_manycore_platform.cpp`: platform-specific tx/rx, initialization

This should make supporting new machines easier. This work would have been needed eventually to support Black Parrot. It has probably already been done by Paul. It was definitely already done by me, for Verilator.

I also took the liberty of removing some old, outdated code. This simplifies the interface between manycore software and platform.

PR #583, opened 2020-05-23
Split platform layer into platform-specific files

drichmond:
On the march toward verilator, I split the VCS and F1-specific code into two new bsg_manycore_platform.cpp files. 

Merge after #579


machines/.gitignore
```diff
@@ -4,3 +4,4 @@ bsg_bladerunner_configuration.rom
 bsg_bladerunner_memsys.rom
 bsg_bladerunner_configuration.v
 bsg_bladerunner_pkg.vh
+bsg_manycore_machine.h
```
drichmond, line 7: Should have been in a previous PR. Let's just ignore that this exists...

testbenches/simlibs.mk
```diff
@@ -127,8 +127,7 @@ include $(TESTBENCH_PATH)/dramsim3.mk
 include $(TESTBENCH_PATH)/infmem.mk
 include $(TESTBENCH_PATH)/libdmamem.mk
 
-SIMLIBS += $(TESTBENCH_PATH)/libfpga_mgmt.so
-SIMLIBS += $(LIBRARIES_PATH)/libbsg_manycore_runtime.so
+SIMLIBS += $(BSG_PLATFORM_PATH)/libbsg_manycore_runtime.so
```
drichmond, line 130: libbsg_manycore_runtime.so in the vcs platform now depends on libfpga_mgmt.so

testbenches/simlibs.mk
```diff
@@ -186,35 +185,6 @@ $(TESTBENCH_PATH)/vcs_simlibs/$(BSG_MACHINE_NAME)/AN.DB: $(BSG_MACHINE_PATH)/syn
 		$(VLOGAN_SOURCES) -f $(TESTBENCH_PATH)/aws.vcs.f \
 		$(VLOGAN_INCLUDES) -l $(BSG_MACHINE_NAME).vlogan.log
 
-# The applications link against the BSG Manycore Libraries, and the FPGA
```
drichmond, line 189: These were moved into the platform-specific library.mk file

libraries/platforms/aws/bsg_manycore_platform.cpp
```diff
@@ -30,328 +21,15 @@
...
         int transmit_vacancy;    //!< Software copy of the transmit vacancy register
-        pci_bar_handle_t handle; //!< pci bar handle
+        int handle; //!< pci bar handle
         hb_mc_manycore_id_t id;  //!< which manycore instance is this
-        uintptr_t      mmio;     //!< pointer to memory mapped io (F1-specific)
+        hb_mc_mmio_t      mmio;     //!< pointer to memory mapped io (F1-specific)
 } hb_mc_platform_t;
 
 
-/* initialize manycore MMIO */
```
drichmond, line 46: These were moved into platform-specific bsg_manycore_mmio.cpp

libraries/bsg_manycore_mmio.h
```diff
@@ -1,74 +0,0 @@
-// Copyright (c) 2019, University of Washington All rights reserved.
-//
```
drichmond, line 2: This file was moved, not deleted

libraries/platforms/aws/bsg_manycore_platform.cpp
```diff
@@ -5,15 +5,6 @@
...
 #include <cstring>
 
-#ifndef COSIM
-#include <fpga_pci.h>
-#include <fpga_mgmt.h>
-#else
-#include <svdpi.h>
-#include <fpga_pci_sv.h>
-#include <utils/sh_dpi_tasks.h>
-#endif
```
drichmond, line 15: These were moved into bsg_manycore_mmio.cpp

libraries/platforms/aws/bsg_manycore_platform.cpp
```diff
@@ -683,25 +361,6 @@ static int hb_mc_platform_get_credits(hb_mc_manycore_t *mc,
...
-}
-
-static void hb_mc_platform_dpi_cleanup(hb_mc_platform_t *pl)
-{
-#ifdef COSIM
-        pl->scope = nullptr;
-#endif
-        return;
-}
-
```
drichmond, line 704: These were moved into MMIO init

libraries/Makefile
```diff
@@ -51,28 +51,37 @@ include $(CL_DIR)/hdk.mk
 # flags, and a list of sources.
 include $(LIBRARIES_PATH)/libraries.mk
 
-build: $(LIBRARIES_PATH)/libbsg_manycore_runtime.so.1.0
+_TARGETS :=
+_DOCSTRING :=
```
drichmond, line 55: Generating documentation strings on the fly, like in baseline

libraries/bsg_manycore.cpp
```diff
@@ -47,16 +46,6 @@
...
-#else
-#include <inttypes>
-#include <stdint.h>
-#include <stdlib.h>
-#include <string.h>
-#include <stdio.h>
-#include <limits.h>
-#include <stdbool.h>
-#include <assert.h>
-#endif
```
drichmond, line 59: Seems unnecessary to keep the old c-specific headers, since we have C++ features throughout

libraries/bsg_manycore.cpp
```diff
@@ -1214,11 +1203,7 @@ int hb_mc_manycore_write_mem(hb_mc_manycore_t *mc, const hb_mc_npa_t *npa,
         if (err != HB_MC_SUCCESS)
                 return err;
 
-        // This pair of matching function calls changes the clock period of the
-        // manycore during data transfer to accelerate simulation
-#ifdef COSIM
-        sv_set_virtual_dip_switch(0, 1);
-#endif
+        hb_mc_platform_start_bulk_transfer(mc);
```
drichmond, line 1206: New function in platform API

libraries/bsg_manycore_platform.h
```diff
@@ -9,70 +10,89 @@
...
+         */
+        int hb_mc_platform_start_bulk_transfer(hb_mc_manycore_t *mc);
+
+        /**
+         * Signal the hardware to end a bulk transfer over the network
+         * @param[in]  mc     A manycore instance initialized with hb_mc_manycore_init()
+         * @return HB_MC_SUCCESS on success. Otherwise an error code defined in bsg_manycore_errno.h.
+         */
+        int hb_mc_platform_finish_bulk_transfer(hb_mc_manycore_t *mc);
+}
```
drichmond, line 97: These are intended to abstract clock switching.

regression/library/test_manycore_init.h
```diff
@@ -27,7 +27,6 @@
 
 #pragma once
 #include <bsg_manycore.h>
-#include <bsg_manycore_mmio.h>
```
drichmond, line 30: No longer available

regression/library/test_rom.h
```diff
@@ -25,10 +25,9 @@
 // (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 // SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-#ifndef TEST_MANYCORE_COMPILE_H
-#define TEST_MANYCORE_COMPILE_H
+#ifndef TEST_ROM_H
+#define TEST_ROM_H
 #include <bsg_manycore.h>
-#include <bsg_manycore_mmio.h>
```
drichmond, line 31: no longer available

drichmond:
This passes regression on timing_v0_8_4 and 4x4_fast_n_fake

drichmond:
@dpetrisko Do you want to look at this too?

README.md
```diff
@@ -34,6 +35,20 @@ to point to any subdirectory of the `machines` directory.
...
+To switch platforms, set the variable `BSG_PLATFORM` in
+[platform.mk](platform.mk). Available platforms are subdirectories in
+[libraries/platforms](libraries/platforms). For the most part, users
+do not need to worry about what platform they are running on;
+`platform.mk` will automatically deduce the correct platform from
+the environment
+
 # Quick-Start
 
 The simplest way to use this project is to clone it's meta-project: [BSG Bladerunner](https://github.com/bespoke-silicon-group/bsg_bladerunner/).
```
dpetrisko, line 54: "its meta-project"

drichmond: I was asking for a code review, not a grammar check..

(JUST KIDDING, I'm glad you found this.)

libraries/bsg_manycore_platform.h
```diff
@@ -9,70 +9,89 @@
...
+         */
+        int hb_mc_platform_fence(hb_mc_manycore_t *mc,
+                                 long timeout);
+
+        /**
+         * Signal the hardware to start a bulk transfer over the network
+         * @param[in]  mc     A manycore instance initialized with hb_mc_manycore_init()
+         * @return HB_MC_SUCCESS on success. Otherwise an error code defined in bsg_manycore_errno.h.
+         */
+        int hb_mc_platform_start_bulk_transfer(hb_mc_manycore_t *mc);
```
dpetrisko, line 87: kind of weird these don't have other parameters? What exactly is the intended implementation here?

README.md
drichmond: Fixed

libraries/bsg_manycore_platform.h
drichmond: This exists to hide the clock switching. Most implementations will just return HB_MC_SUCCESS

drichmond: I meant to add a comment about that. Will do

dpetrisko:
The nomenclature here for what is a "platform" is a little confusing.  For instance, a BP+hammerblade system should run on Verilator or VCS.  So BP will implement the platform interface in cpp, but the system is actually running "on" VCS.  Not sure what the solution is, just thinking out loud

README.md
dpetrisko: :)

drichmond:
> The nomenclature here for what is a "platform" is a little confusing. For instance, a BP+hammerblade system should run on Verilator or VCS. So BP will implement the platform interface in cpp, but the system is actually running "on" VCS. Not sure what the solution is, just thinking out loud

Yeah. I agree. I haven't come up with a better alternative so I'm open to suggestions. would `aws-vcs` be acceptable?

libraries/bsg_manycore_platform.h
drichmond: Fixed

drichmond:
> Could we make the interface for this function more generic? It's basically an NPA memcpy right?

Which?

dpetrisko:
> Could we make the interface for this function more generic? It's basically an NPA memcpy right?
> 
> Which?

Sorry, the clock-switching ones

dpetrisko:
> Yeah. I agree. I haven't come up with a better alternative so I'm open to suggestions. would aws-vcs be acceptable?

<host>-<platform> tuples sound good to me.  blackparrot-vcs, aws-f1, etc...

drichmond:
> > Yeah. I agree. I haven't come up with a better alternative so I'm open to suggestions. would aws-vcs be acceptable?
> 
> host-platform tuples sound good to me. blackparrot-vcs, aws-f1, etc...

Can do!

machines/.gitignore
mrutt92: No - separate PR! :-p

mrutt92:
You know... start/stop bulk transfer are only hacks because the comment admits that they're hacks. Abstractly they are "Hooks that the platform may implement to perform special initialization and cleanup tasks for assisting in doing bulk transfers from the host and device." See - now it's not a hack.

libraries/bsg_manycore.cpp
mrutt92: This has always been 100% unnecessary because this file is C++ source file and will never be compiled without __cplusplus being defined. The only place this maybe makes sense is in header files that might be included by C client source code.

libraries/bsg_manycore.cpp
mrutt92: nice

libraries/bsg_manycore_platform.h
mrutt92: nice

libraries/Makefile
```diff
@@ -51,28 +51,37 @@ include $(CL_DIR)/hdk.mk
...
-	mv $(notdir $<) /usr/lib64/
-	ln -sf /usr/lib64/$(notdir $<) /usr/lib64/libbsg_manycore_runtime.so.1
-	ln -sf /usr/lib64/$(notdir $<) /usr/lib64/libbsg_manycore_runtime.so
-	cp -t /usr/include $(LIB_HEADERS)
+_DOCSTRING += "Rules from Makefile\n"
+_TARGETS :=
 
+.DEFAULT_GOAL := help
+_TARGETS += "help"
+_DOCSTRING += "    help: \n"
```
mrutt92, line 62: You know - you could do this kind of thing like this:

```
_DOCSTRING-atargetname := "Help message for a target name"
targets += atargetname
```

then in help you do:
```
help:
    @(foreach t, $(targets), echo "$t:"; echo "    $(_DOCSTRING-$t)";)
```
True story - that works... and then you don't need to report the formatting of your help message everywhere you add a new target.

Not saying that you adopt that strategy here but maybe something to think about in the future.

drichmond:
Clever. I like it. 

One downside, though, is that I can’t document where the rules are coming from as easily. Yes, I could add it to every rule but that seems verbose.

But it’s a great and clever suggestion

mrutt92:
yeah well... where are the platform specific functions for vcs?

drichmond:
I managed to localize them to the MMIO and reuse the AWS platform layer.

drichmond:
Remember how we were talking about duplicate code? This is how I solved that Issue

libraries/platforms/vcs/library.mk
```diff
@@ -0,0 +1,73 @@
...
+# ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+ifndef VCS_HOME
+$(error $(shell echo -e "$(RED)BSG MAKE ERROR: VCS_HOME environment variable undefined. Are you sure vcs-mx is installed?$(NC)"))
+endif
+
+PLATFORM_CXXSOURCES += $(BSG_F1_DIR)/libraries/platforms/vcs/bsg_manycore_mmio.cpp
+PLATFORM_CXXSOURCES += $(BSG_F1_DIR)/libraries/platforms/aws/bsg_manycore_platform.cpp
```
mrutt92, line 33: put a comment here to draw some attention to the fact that your are borrowing code from another platform.

mrutt92:
I see - so in the makefile for the vcs platform you 'borrow' the platform.cpp that's in the aws platform. I feel like that's worth spelling out a little more in either comment or README form.

mrutt92:
Regarding your response to my help target suggestion: I don't understand the downside you raise (maybe something about the order in your list or something?) but I also don't feel strongly enough to insist you do it my way. You can disregard it with respect to me approving this PR.

PR #587, opened 2020-05-30
Adds a machine header which is used by test_rom.c

mrutt92:
* Fixes broken test `test_rom`

### DIFF
--- a/testbenches/simlibs.mk
+++ b/testbenches/simlibs.mk
@@ -199,7 +199,6 @@ $(LIB_OBJECTS): CFLAGS   += -DCOSIM
 
 # libfpga_mgmt will be compiled in $(TESTBENCH_PATH), so direct the linker there
 $(LIB_OBJECTS): INCLUDES += -I$(VCS_HOME)/linux64/lib/
-$(LIB_OBJECTS): $(BSG_MACHINE_PATH)/bsg_manycore_machine.h
 $(LIBRARIES_PATH)/libbsg_manycore_runtime.so.1.0: LDFLAGS +=-L$(TESTBENCH_PATH) 
 $(LIBRARIES_PATH)/libbsg_manycore_runtime.so.1.0: LDFLAGS +=-Wl,-rpath=$(TESTBENCH_PATH)
 $(LIBRARIES_PATH)/libbsg_manycore_runtime.so.1.0: $(TESTBENCH_PATH)/libfpga_mgmt.so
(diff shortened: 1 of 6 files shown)


testbenches/simlibs.mk
```diff
@@ -199,6 +199,7 @@ $(LIB_OBJECTS): CFLAGS   += -DCOSIM
 
 # libfpga_mgmt will be compiled in $(TESTBENCH_PATH), so direct the linker there
 $(LIB_OBJECTS): INCLUDES += -I$(VCS_HOME)/linux64/lib/
+$(LIB_OBJECTS): $(BSG_MACHINE_PATH)/bsg_manycore_machine.h
```
drichmond, line 202: Why here?

mrutt92: "If you don't think too good, don't think too much."
Where do you think it should go?

mrutt92: Oh duh

mrutt92: `libraries.mk`

drichmond: You really are a child of the northeast

drichmond: I had to google that reference

drichmond: I'd like to understand why bsg_manycore_machine.h is a dependency of the build library, when it's only included in the test_rom.cpp source

drichmond: I think you've done the right thing here (make a header)

mrutt92: I made that choice because I didn't want every test that might use it to need to explicitly list it as a dependency.

drichmond: I need to think about this more. I don't like the concept of depending on a file that isn't part of the compiled source. I'm also concerned that this introduces issues when switching machines.

We've resisted creating a machine.h file in the past so that we don't bake-in machine configurations into the library. This opens that up as a possibility.

mrutt92: The way I see it; we're giving people a tool they can use when it's appropriate, as is the case for `test_rom.c`.

For example: `std::map` is a great tool that can be used the wrong way. That doesn't make it less of a good thing.

This can be a nice thing to have for kernels - there might be value in knowing how many banks make up your L2 and how large the lines and what the stripe size is.

drichmond: OK, yes. I can't guarantee that I'll keep this particular rule long term, but OK.

PR #574, opened 2020-04-29
WIP: Verilator

drichmond:
This PR adds support for Verilator simulation. 

It depends on PRs:

[274](https://github.com/bespoke-silicon-group/basejump_stl/pull/274) in BaseJump
[279](https://github.com/bespoke-silicon-group/basejump_stl/pull/279) in BaseJump

(These PRs are combined in the [manycore_verilator](https://github.com/bespoke-silicon-group/basejump_stl/tree/manycore_verilator) branch)

[295](https://github.com/bespoke-silicon-group/bsg_manycore/pull/295) in Manycore
[307](https://github.com/bespoke-silicon-group/bsg_manycore/pull/307) in Manycore

(These PRs are combined in the [verilator](https://github.com/bespoke-silicon-group/bsg_manycore/tree/verilator) branch)

At the moment, this PR does not support generating waveforms (verilator generates VCD files, which is inefficient), or profiler outputs. 

The profilers are not supported because `forever \n @posedge clk` statements are not supported in Verilator. When that is fixed, these can be re-added.

You can run verilator simulation using the [verilator](https://github.com/bespoke-silicon-group/bsg_bladerunner/tree/verilator) branch in Bladerunner

This PR passes regression on 4x4_fast_n_fake

(EDIT: This PR passes, but it does NOT support multiple init/cleanup runs -- so two regression tests have to be cut out)

PR #591, opened 2020-06-12
Verilator (Second attempt)

drichmond:
Added support for Verilator, with new dpi-verilator platform. This PR basically does the same work as #574, except that the changes are minimal since all the other changes were split out into other PRs. 

* Created link.mk, hardware.mk, and library.mk for verilator
* Added bsg_manycore_platform.cpp for dpi-verilator (incomplete)
* Added generic interface bsg_manycore_verilator.cpp/hpp to separate Verilator compilation from runtime compilation.
* Added verilator top-level simulation module

This currently passes regression on 4x4_fast_n_fake and timing_v0_8_4 with master of basejump, and bsg_manycore. 

It should be merged after #598 (which is a small PR)


drichmond:
Superseded by #610

PR #598, opened 2020-06-17
Added a check in platform library to prevent reused IDs

drichmond:
* Updated regression test

Fails regression before (without updated library), passes after

### DIFF
--- a/libraries/platforms/aws-fpga/bsg_manycore_platform.cpp
+++ b/libraries/platforms/aws-fpga/bsg_manycore_platform.cpp
@@ -4,6 +4,7 @@
 #include <bsg_manycore_printing.h>
 
 #include <cstring>
+#include <set>
 
 /* these are convenience macros that are only good for one line prints */
 #define platform_pr_dbg(m, fmt, ...)                    \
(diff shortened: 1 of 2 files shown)


libraries/platforms/aws-fpga/bsg_manycore_platform.cpp
```diff
@@ -361,6 +362,9 @@ static int hb_mc_platform_get_credits(hb_mc_manycore_t *mc,
 
 
 
+// This track active manycore machine IDs
+static std::set<hb_mc_manycore_id_t> active_ids;
```
mrutt92, line 366: I'm not sure I would implement this at the platform level because now all platforms will need to implement code that looks quite a bit like this. Implementing this at the manycore level would reduce code duplication.

drichmond: Max and I had a long friday afternoon debate about this. 

The answer is yes. This is the right way to do things, but it would require some re-engineering. Since this is small potatoes compared to everything else I want/need to do I'm going to merge this.

However, we can file an issue if anyone feels strongly

dpetrisko: OOC what's a machine id?

drichmond: A machine ID was originally intended for identifying different AWS FPGAs on an instance with multiple FPGAs. 

So, different, unique instances of a manycore managed by the runtime. 

In Verilator, it could be multiple independent instances of simulation (though I haven't tested it and explicitly disallow it)

Does that answer your question?

PR #601, opened 2020-06-26
Added cycle counter interface in HDL and in Software

drichmond:
The cycle counter is a platform level feature, since each platform will implement it differently. This implementation is for AWS (both simulation and emulation). The cycle counter can be read by the host at any time, and is located in the AXI to MCL Interface.

Verilator also supports this, but it is in a different branch.

At the platform level, each platform is required to implement hb_mc_platform_get_cycle. This is exposed at the manycore level through hb_mc_manycore_get_cycle (with the same API). 

I have not exposed this at the CUDA level. It would be possible to call get_cycle after a kernel terminates, but this will cause a small (20 cycle?) latency between when the packet is seen in the fifo, and when get_cycle is read (~20 cycles to read a packet out, in simulation). 

I've added an example/test that demonstrates the use of this API. 

Max for review. Lin for input.

### DIFF
--- a/examples/library/test_get_cycle.cpp
+++ b/examples/library/test_get_cycle.cpp
@@ -81,7 +81,7 @@ int test_get_cycle (int argc, char **argv) {
                         goto cleanup;
                 }
 
-                if(last >= cycle){
+                if(last > cycle){
                         bsg_pr_test_err("Cycle counter not increasing!\n");
                         fail = HB_MC_FAIL;
                         goto cleanup;
--- a/examples/library/test_get_cycle.cpp
+++ b/examples/library/test_get_cycle.cpp
@@ -62,6 +62,15 @@ int test_get_cycle (int argc, char **argv) {
         uint32_t dram_coord_y = hb_mc_config_get_dram_y(config);
         hb_mc_npa_t npa = { .x = dram_coord_x, .y = dram_coord_y, .epa = 0};
 
+        bsg_pr_test_info("Testing invalid input (This will print an ERROR message)\n");
+        rc = hb_mc_manycore_get_cycle(&mc, nullptr);
+        if(rc != HB_MC_INVALID){
+                bsg_pr_test_err("hb_mc_manycore_get_cycle() failed to fail on invalid input");
+                fail = HB_MC_FAIL;
+                goto cleanup;
+        }
+                
+
         last = 0;
         for(int i = 0 ; i < 10; ++i){
                 rc = hb_mc_manycore_get_cycle(&mc, &cycle);
--- a/examples/library/test_get_cycle.cpp
+++ b/examples/library/test_get_cycle.cpp
@@ -1,0 +1,110 @@
+// Copyright (c) 2019, University of Washington All rights reserved.
+// 
+// Redistribution and use in source and binary forms, with or without modification,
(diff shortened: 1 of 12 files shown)


drichmond:
@vb000 I'd like your review, too. 🚀

libraries/platforms/aws-fpga/bsg_manycore_platform.cpp
```diff
@@ -488,3 +488,38 @@ int hb_mc_platform_fence(hb_mc_manycore_t *mc,
...
 
+
+/**
+ * Get the current cycle counter of the Manycore Platform
+ *
+ * @param[in]  mc     A manycore instance initialized with hb_mc_manycore_init()
+ * @param[out] time   The current counter value.
+ * @return HB_MC_SUCCESS on success. Otherwise an error code defined in bsg_manycore_errno.h.
+ */
+int hb_mc_platform_get_cycle(hb_mc_manycore_t *mc, uint64_t *time)
```
vb000, line 499: Minor... may be `time` could be a reference to avoid null pointers?

drichmond: Yep. Good idea

drichmond:
Yeah, or you can think of this as the version of `bsg_time` that people should actually use.

vb000:
> Yeah, or you can think of this as the version of `bsg_time` that people should actually use.

Should remove `bsg_time` or name this `bsg_time` then? This could remove some `ifdef`s to make the host code cleaner...

drichmond:
It's gone in the verilator PR. I can separate into a separate PR.

libraries/platforms/aws-fpga/bsg_manycore_platform.cpp
drichmond: Added a check in bsg_manycore.cpp (since that is the only place this file gets used)

drichmond: Also added a check in the regression test.

dpetrisko:
I think there’s value in both get_cycle and get_time. RISC-V platforms typically have a RTC component. If a host doesn’t support an RTC, then get_time can just call get_cycle

drichmond:
@dpetrisko I'm not sure I follow. This is for the host interface

drichmond:
bsg_time is/was a wrapper around a DPI call that returned the current simulation time. This is a formalized way to get the cycle counter on the manycore.

dpetrisko:
Yeah I’m saying a host such as BP can support calls both cycle count and wall-clock times. Both may be valuable, since clock frequency will be different between the platforms.

examples/library/test_get_cycle.cpp
```diff
@@ -0,0 +1,119 @@
...
+        for(int i = 0 ; i < 10; ++i){
+                rc = hb_mc_manycore_get_cycle(&mc, &cycle);
+                if(rc != HB_MC_SUCCESS){
+                        bsg_pr_test_err("Failed to read cycle counter: %s\n",
+                                        hb_mc_strerror(rc));
+                        fail = rc;
+                        goto cleanup;
+                }
+
+                if(last >= cycle){
```
mrutt92, line 84: make this > not >=

Issue #339, opened 2019-07-10
Print cycles/second when cosimulation terminates


drichmond:

taylor-bsg:
Cycles per wall clock second? Or..?

drichmond:
My intent was cycles per wall clock second. But, I'm open to other suggestions. We could also print it out at an arbitrary interval so that we get a trace of when cosimulation speeds up and slows down.

taylor-bsg:
Maybe just wall clock time? I don’t know how cycles per wall second would
be useful?

drichmond:
I thought you asked for this in the past. At the very least, this was requested in the meeting yesterday.

Issue #315, opened 2019-06-26
This is check is going to be too forgiving in no DRAM mode


max-ruttenberg:
https://github.com/bespoke-silicon-group/bsg_f1/blob/726ede5ee3c1d85b7cf0dada676127777b0d1b3d/cl_manycore/libraries/bsg_manycore_eva.cpp#L319

drichmond:
@max-ruttenberg  how so? Maybe related to #333?

drichmond:
is this fixed?

max-ruttenberg:
I honestly don't know. I was way too vague when I posted this issue. Let's close it.

PR #603, opened 2020-06-30
Added profiler API

drichmond:
* Added get_icount methods
* Added test_profiler regression test (and example)

This PR adds support for querying device execution statistics at runtime, and a regression test/example to demonstrate what good performance looks like.

This is the output of test_profiler: 

```
INFO:    
INFO:    
INFO:    ====== EXECUTION STATISTICS ====== 
INFO:    Cycles: 25163
INFO:    Instrs: 24213, Flop Count: 11968
INFO:    Flop Rate: 49.43
INFO:    IPC: 0.96
INFO:    ====== END EXECUTION STATISTICS ====== 
INFO:    
INFO: 
```

Users can query three types of instructions: `e_instr_all` for all instructions executed, `e_instr_float` for floating point instructions, `e_instr_integer` for integer instructions.

Depends on [#340](https://github.com/bespoke-silicon-group/bsg_manycore/pull/340) and [#339](https://github.com/bespoke-silicon-group/bsg_manycore/pull/339) and [#315](https://github.com/bespoke-silicon-group/basejump_stl/pull/315)


drichmond:
This passes in 4x4_fast_n_fake and timing_v0_8_4

examples/cuda/test_profiler.cpp
```diff
@@ -0,0 +1,343 @@
+// Copyright (c) 2020, University of Washington All rights reserved.
```
vb000, line 1: I was looking for an example to initialize the profiler. I'd imagine this file might've to do that, but couldn't a call that does that... how does this test initialize profiler?

drichmond: It happens internally: https://github.com/bespoke-silicon-group/bsg_replicant/blob/18d4ccdcc88dbc6e482bef2bf318bcf87f799dbc/libraries/platforms/aws-fpga/bsg_manycore_platform.cpp#L458

libraries/features/profiler/bsg_manycore_profiler.hpp
```diff
@@ -0,0 +1,43 @@
+#ifndef __BSG_MANYCORE_PROFILER_HPP
+#define __BSG_MANYCORE_PROFILER_HPP
+#include <bsg_manycore.h>
+#include <string>
+
+// Since the definition of hb_mc_profiler_t is implementation
+// dependent, we use void *
+typedef void* hb_mc_profiler_t;
```
vb000, line 8: This is interesting. Where are the actual contents defined in a platform usually?

drichmond: Are you asking, what does the void * point to? 

In a simulation platform, it points to a std::vector: 

https://github.com/bespoke-silicon-group/bsg_replicant/blob/18d4ccdcc88dbc6e482bef2bf318bcf87f799dbc/libraries/features/profiler/simulation/bsg_manycore_profiler.cpp#L58

I used a void * so that another platform that wants to implement a profiler could design its own implementation.

examples/cuda/test_profiler.cpp
vb000: Okay. So, turned on by default?

drichmond: Yeah. It is already. This hooks into the `vanilla_core_profiler` module, not `vanilla_core_trace`

libraries/features/profiler/bsg_manycore_profiler.hpp
vb000: I makes sense. That, host code wouldn't have to worry about managing profiler but can just get info by calling get_icount etc. at will.

examples/cuda/test_profiler.cpp
vb000: I makes sense. That, host code wouldn't have to worry about managing profiler but can just get info by calling get_icount etc. at will.

drichmond: Right. And if it isn't implemented it'll return NO_IMPL. 

Though my motivation was somewhat more mundane: It avoids `ifdef COSIM`

libraries/features/profiler/bsg_manycore_profiler.hpp
vb000: Great! The details of the contents of void* are abstracted by api functions (like get_icount and get_cycles)?

examples/cuda/test_profiler.cpp
vb000: Since this profiler actual hardware, there's no benefit to stopping it also ;)

libraries/features/profiler/bsg_manycore_profiler.hpp
drichmond: Yep - https://github.com/bespoke-silicon-group/bsg_replicant/blob/18d4ccdcc88dbc6e482bef2bf318bcf87f799dbc/examples/cuda/test_vec_add.c#L171

(Something didn't get pushed up, those should be enum values not raw integers)

PR #606, opened 2020-07-06
Add tracer C/C++ API

drichmond:
Depends on [346](https://github.com/bespoke-silicon-group/bsg_manycore/pull/346/files)

This adds an API for enabling/disabling trace file and log file generation. This can drastically reduce runtime and trace file size when datasets are large. 

This is necessary because parsing "plus args", i.e. `+trace` isn't clean in Verilator.

The four new functions are
`bsg_manycore_trace_enable/disable` and `bsg_manycore_log_enable/disable`. The former creates `vanilla_operation_trace.csv` and the latter generates `vanilla.log`

Also added a regression test/example.

### DIFF
--- a/examples/cuda/Makefile
+++ b/examples/cuda/Makefile
@@ -96,6 +96,7 @@ INDEPENDENT_TESTS += test_softmax
 INDEPENDENT_TESTS += test_log_softmax
 INDEPENDENT_TESTS += test_hammer_cache
 INDEPENDENT_TESTS += test_profiler
+INDEPENDENT_TESTS += test_tracer
 # Tests currently not working with new FPU.
 #INDEPENDENT_TESTS += test_conv1d
 #INDEPENDENT_TESTS += test_conv2d
--- a/examples/cuda/test_tracer.cpp
+++ b/examples/cuda/test_tracer.cpp
@@ -1,0 +1,250 @@
+// Copyright (c) 2019, University of Washington All rights reserved.
+// 
+// Redistribution and use in source and binary forms, with or without modification,
+// are permitted provided that the following conditions are met:
+// 
+// Redistributions of source code must retain the above copyright notice, this list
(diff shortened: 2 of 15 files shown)


dpetrisko:
plusargs should work in verilator. What issue are you running into?

dpetrisko:
(this is still a great change for trace size reduction, just wondering if there's a Verilator bug to be discovered :))

drichmond:
They do!

I said it wasn't clean, not that it doesn't work. Let me clarify.

We run argument parsers in our programs, and they don't like when additional, non-recognized arguments are passed in (e.g. +trace). My impression is that in order to prevent this you would have to run verilator's arg parser first, and then run our arg parser. 

Ideally we could hide the Verilator arg parser in manycore init, but we run our arg parser before manycore init. 

So my impression is that the solution is requiring the user to run the verilator arg parser in main when they're running in verilator. I think this exposes too many platform details to the user, so I came up with a way around it. 

The issue, and the reason I say this isn't clean is because it exposes the user to the implementation.

drichmond:
In short, writing the trace file is slow. 

This PR provides the ability to select WHEN the trace file is being written. For example, we don't usually care about the data in the trace file when DRAM is being loaded, or the icache is being initialized. That's just overhead, and writing the trace file slows down simulation during those times.

We normally only care when the kernel starts executing. This API allows the user to start trace generation immediately before the kernel starts executing and stop trace generation immediately after. 

That speeds up simulation. It also decreases the trace file size.

drichmond:
When #606 is merged, this is ready for review.

drichmond:
@bornaehsani Did the generated traces/logs parse correctly?

bornaehsani:
> @bornaehsani Did the generated traces/logs parse correctly?

Yes stats, pc_histogram, vanilla core blood graphs and vcache trace graphs generated!

PR #325, opened 2019-07-03
Manycore FPU Changes

drichmond:
Depends on a tapeout-enabled branch in bsg_manycore (Use bladerunner to get corresponding branches)

This is technically four PRs:

* FPU features (Simulation and Compilation File Lists)
* Support for no-dram mode (Lower-Level and CUDA API)
* Early support for "Machine-based" linker script changes
* Support for @tommydcjung's trace API in VCS

Issue #324, opened 2019-07-02
Rewrite configuration init to handle no-dram mode


drichmond:

bornaehsani:
We also need to set the correct dram size in both dram/no-dram modes, when the manycore hb_mc_config_t struct is initialized. The cuda memory allocator should throw an error if it runs out of space (which it doesn't right now, causing memory locations to get overwritten and tiles to run with wrong symbol values  and spin out of control with invalid EVA accesses)

https://github.com/bespoke-silicon-group/bsg_f1/blob/master/cl_manycore/libraries/bsg_manycore_cuda.cpp#L1295

drichmond:
Related to #333

drichmond:
@bornaehsani I think this is related to our discussion yesterday. Yes?

bornaehsani:
Very much yes, if we include a parameter dram/no-dram inside ROM, the CUDA API can call hb_mc_disable_dram inside device_init, which means we will not need two versions of each regression tests.

max-ruttenberg:
We were discussing this earlier. What was the conclusion?

drichmond:
@bornaehsani @mrutt92 I think this is solved, right?

PR #610, opened 2020-07-14
Verilator

drichmond:
Third (and final?) attempt

This adds support for verilator. It passes regression on the timing and fake machines. It supports trace, profilers, and logging.


drichmond:
This does not pass regression on the crossbar machines. A packet is getting dropped somewhere. @tommydcjung, Can you help me out 1:1? There are no debug messages being generated.

drichmond:
This PR passes regression on timing machines, and fake machines. 

It does not pass regression on the baseline/crossbar machines, because a number of `cuda` tests hang (or take a long time to execute). 

The behavior is... odd. 

- The hangs only occur in the `cuda` suite, but do not occur in the `spmd` or `library` test suite
- The ones that hang have barriers, but other code with barriers passes (e.g. test_print_tile_info)
- There are no errors or warnings during simulation

I'm not sure this is a barrier issue, but I could be wrong. The same machines pass in VCS, and the same kernels pass on 4x4_fast_n_fake. 

To reproduce this issue, checkout this branch and master of manycore, and basejump. I've modified baseline_v0_8_4 to make it a 4x4 machine (temporarily) and simulate faster. 

The tests listed below will generate a vcd file (waveform) that can be analyzed in DVE. They also generate the standard log files. 

With test_vec_add_parallel, it generates a 2.1 GB vcd file. To generate it yourself, check out this branch, and run `make test_vec_add_parallel` (you must have bladerunner with Verilator)

Or, you can get it from me

The following tests currently hang: 

```
test_softmax
test_log_softmax
test_matrix_mul_shared_mem
test_shared_mem_load_store
test_vec_add_dma
test_float_vec_mul
test_float_matrix_mul_shared_mem
test_tracer
test_float_vec_add
test_float_vec_sqrt
test_float_all_ops
test_float_vec_exp
test_float_matrix_mul
test_float_vec_div
test_vec_add_shared_mem
test_vec_add_parallel
test_vec_add_parallel_multi_grid
test_float_vec_log
test_matrix_mul
test_vec_add
test_vec_add_serial_multi_grid
```

taylor-bsg:
Interesting .. it could be exposing synchronization bugs in the software,
for example the bug in the barrier that borna just found where it was not
doing a fence before entering the barrier...

drichmond:
> Interesting .. it could be exposing synchronization bugs in the software, for example the bug in the barrier that borna just found where it was not doing a fence before entering the barrier...

I tried that originally, and just tried again. So far, no dice. 

My very initial impression is that something is going funky during kernel launch. The first line of the kernel is a `lui` and the next cycle is a dram remote load stall. I can show in the meeting tomorrow. 


What causes this? I don't know. One really difficult aspect of verilator is the lack of Z's/X's.

drichmond:
For example, I spent part of yesterday, and half of today, debugging what would have been a Z/X in vcs. 

It's a good argument for getting this top-level working in VCS.

taylor-bsg:
Yeah I think we should always start with VCS for this reason. does it pass
with VCS?

m

drichmond:
Yep!

drichmond:
Different top level design, though.

dpetrisko:
> What causes this? I don't know. One really difficult aspect of verilator is the lack of Z's/X's.

One smoke test is trying verilator with --x-initial 0/1, --x-assign 0/1.  If you get different behavior, you'll know it's (probably) x-prop.

drichmond:
Good News: I got it to work and pass most regression tests.
Bad News: There's still an issue.

@vb000 and I looked into this after the manycore meeting. With his help, I was able to narrow down the issue to a WAW hazard detected by the scoreboard.

What seems to be happening is that a DRAM load is being issued to load a CUDA argument. This CUDA argument isn't needed by the kernel, so the compiler overwrites that register (with a `lui`). During execution the scoreboard detects a WAW hazard (`rd_depend_on_sb`) because the remote load is still in flight and stalls the pipeline. The remote load returns (out of order?) and gets stuck behind another load, and isn't read. 

That's my interpretation of what is happening. 

What I did was insert a bunch of `nop`s into the assembly for loading arguments. This changes timing (enough) so that the dependency is resolved by the time the register is written. Obviously this isn't a fix, but it does demonstrate that the issue is a race condition. 

I can create VCD files for anyone interested. I have a waveform open at the moment. 

@tommydcjung, your thoughts?

libraries/platforms/dpi-verilator/bsg_manycore_platform.cpp
```diff
@@ -0,0 +1,553 @@
...
+{
+
+        int r = HB_MC_FAIL, err;
+        
+        hb_mc_platform_t *platform = new hb_mc_platform_t;
+        std::string hierarchy = "TOP.manycore_tb_top";
+        hb_mc_idx_t x, y;
+        hb_mc_config_raw_t rd;
+
+        // Uncomment THIS and the statement in verilator_top.sv to enable tracing
```
dpetrisko, line 192: Maybe a #define to enable tracing?

drichmond: I thought about this -- especially after the last couple of days where Tommy and I have been knee-deep in vcd files.

Yes, you're right. That would be the way to make it available. 

My opinion is that I don't want people to use verilator for waveform debugging, unless they're an expert. The amount of weird odds and ends that occur without Z/X and the size of vcd files make it pretty unfriendly. I want to make users have to overcome a hurdle to use it. 

I'm open to changing it. I'm also open to creating a .debug binary build option (perhaps another PR) that build the waveform generating executable.

dpetrisko: sounds like another PR :)

libraries/platforms/dpi-verilator/hardware/verilator_top.sv
```diff
@@ -0,0 +1,779 @@
+module manycore_tb_top
```
tommydcjung, line 1: Instead of creating another top-level, I would suggest refactoring cl_manycore.sv to factor all the "hardware", which has the ports only for clk and reset. That could have two wrappers for vcs and verilator each, where the simulator specific things gets instantiated. Otherwise, this will become a maintainability issue. We want to make sure that things are structured in an efficient way from the beginning so that future technical debts are minimized...

dpetrisko: Yep, agree completely. This is the approach we take in BP:

https://github.com/black-parrot/black-parrot/blob/master/bp_top/test/tb/bp_tethered/testbench.v

https://github.com/black-parrot/black-parrot/blob/master/bp_top/test/tb/bp_tethered/test_bp.v

https://github.com/black-parrot/black-parrot/blob/master/bp_top/test/tb/bp_tethered/test_bp.cpp

drichmond: I agree.

However, I can't refactor `cl_manycore.sv`. The interface for that module is defined by AWS. 

We could refactor the manycore hierarchy that we reuse between the two modules. I see value there, and I think this what you are describing.

I'd like to take it a step farther - I think spmd_testbench should be included as well.

I don't believe that should be part of this PR.

drichmond: I do think `cl_manycore.sv` is due for a refactor. And I support the motivation of stomping out technical debt/bit-rot and easing maintenance. 

So yes, I wholeheartedly agree with what you're both saying. In case that wasn't clear

drichmond:
With [#324](https://github.com/bespoke-silicon-group/basejump_stl/pull/324) and [#325](https://github.com/bespoke-silicon-group/basejump_stl/pull/325) this PR passes regression with Verilator on 4x4_fast_n_fake, timing_v0_8_4, baseline_v0_8_4. It also passes VCS regression on 4x4_vcache_blocking_f1_model, 4x4_fast_n_fake, timing_v0_8_4, and baseline_v0_8_4.

(Both of those PRs have been approved, just waiting for confirmation from @taylor-bsg)

Something to note for later: VCS and Verilator act very differently. Some tests that are short in VCS take a long time in verilator (`test_vcache_flush`). Some tests that are short in Verilator take longer in VCS. My THEORY is that Verilator simulates arithmetic operations very well, but does not simulate the network well. Something to look into, when we want to optimize Verilator performance. 

Thanks to Dan, Tommy and Bandhav for their help over the last couple of days.

PR #628, opened 2020-07-24
Add dpi-vcs platform

drichmond:
Merge/rebase after #627. This PR looks complex, but it is not and it will be simple.

This PR adds the `dpi-vcs` platform, which uses the same C/C++ libraries and RTL as the `dpi-verilator` platform but simulates in VCS.

Passes regression on 

`dpi-vcs`: 4x4_fast_n_fake, timing_v0_8_4, baseline_v0_8_4


dpetrisko:
LGTM!

PR #629, opened 2020-07-27
Program Init DMA

vb000:
Updated program init to use DMA for DRAM only segments.


drichmond:
Do you know if this makes short-running tests, take longer?

vb000:
Didn't measure it, but bsg_replicant's regression was as usual, quick with 32 threads. Do you have any particular tests in mind that I should measure?

libraries/bsg_manycore_loader.cpp
```diff
@@ -179,13 +179,29 @@ static int hb_mc_loader_eva_write(const Elf32_Phdr *phdr,
...
+                           __func__);
+                return rc;
+        }
+
+        if (hb_mc_manycore_npa_is_dram(mc, &start_npa)) {
+                rc = hb_mc_manycore_eva_write_dma(mc, map, &tile,
+                                                  &start_eva, data, sz);
+                if (rc == HB_MC_SUCCESS) {
+                        return rc;
+                }
```
drichmond, line 201: There's a larger space than not HB_MC_SUCCESS here. It can FAIL, or it can be UNSUPPORTED. Unsupported is fine, but FAIL is not -- the device could be in an unstable state.

vb000:
> Didn't measure it, but bsg_replicant's regression was as usual, quick with 32 threads. Do you have any particular tests in mind that I should measure?

But I did measure the run times of a dummy pytorch test, as pytorch binary is of significant size. Runtimes with and without this change are 48.7s and 16.7s, respectively. The test involved loading binary, creating and copying a tiny tensor.

drichmond:
There's definitely benefits here. I want to understand the costs, too.

vb000:
Okay, I'll time the regression. Why do you suspect this would make short test slower?

drichmond:
I suspect it could, but don't know. That's why I'm asking.

Did you only test it on fastnfake?

vb000:
I only measure on fastnfake. But I also used it on timing_v0_16_8.

drichmond:
You need to run regression on both

libraries/bsg_manycore_loader.cpp
vb000: I was little unsure what are the possibilities here. Do you know where I could find the complete set of outcomes of a DMA call?

drichmond: If this documentation is insufficient, then it should be updated: 

https://github.com/bespoke-silicon-group/bsg_replicant/blob/029602cf3c6dabbf14c053334c3dca731bc51b72/libraries/bsg_manycore.h#L311

drichmond: (By me, or max, or someone else)

libraries/bsg_manycore.h
```diff
@@ -293,6 +293,19 @@ extern "C" {
         /* DMA API */
         /***********/
 
+        /**
```
mrutt92, line 296: Not really sure what changed here.

mrutt92: Oh you removed static - purposefully? I would put it back.

libraries/bsg_manycore_loader.cpp
mrutt92: That comment basically covers it... you should assume everything can also return FAIL.

mrutt92: That comment basically covers it... you should assume everything can also return FAIL.

mrutt92:
One cost would be if there was a program with large test that didn't fit in icache.

libraries/bsg_manycore_loader.cpp
```diff
@@ -179,13 +179,30 @@ static int hb_mc_loader_eva_write(const Elf32_Phdr *phdr,
...
+                bsg_pr_err("%s: Failed to translate EVA into a NPA\n",
+                           __func__);
+                return rc;
+        }
+
+        if (hb_mc_manycore_npa_is_dram(mc, &start_npa) &&
+            hb_mc_manycore_dram_is_enabled(mc)) {
+                rc = hb_mc_manycore_eva_write_dma(mc, map, &tile,
+                                                  &start_eva, data, sz);
+                if (rc != HB_MC_NOIMPL) {
```
mrutt92, line 200: I think this needs to be an if-else.

libraries/bsg_manycore.h
vb000: This is a nice api function to check if npa is a dram npa. I wanted to check that before invoking DMA. Do we need this to be static?

libraries/bsg_manycore_loader.cpp
vb000: We return either way, either success or error, expect for NOIMPL, in which case, we fall back. So, else is not needed...?

libraries/bsg_manycore.h
mrutt92: It's defined in a header outside of a class declaration - it needs to be static otherwise you can run into linker errors.

libraries/bsg_manycore_loader.cpp
mrutt92: But what if you succeed?

mrutt92: This code says if you get success you will also fall through and call eva_write.

mrutt92: Oh I see. NM. Confusing code style though - you should add a comment if nothing else.

libraries/bsg_manycore.h
vb000: Oh the linker error, yes. Done!

vb000:
> One cost would be if there was a program with large text that didn't fit in icache.

Trying to understand more, the major cost of DMA usually is vcache invalidation, right? That's absent in this case, so this should strictly be faster than no DMA? Or may be I'm missing some details relating vcache and program loading.

drichmond:
Let's take an example where: 

- The program does not fit in icache. 
- The program fits in the column caches

(it would be trivial to construct a case where this happens)

With DMA, the column caches are not warmed. It could be possible for a simple program to not only icache miss, but icache and column cache miss. In addition, the host will have to invalidate all of the entries in the column caches. This is not a simple operation. 

This could cause a simple program to run slower.

In addition, this could cause a bug. It shouldn't but, hey, we've seen stranger things.

mrutt92:
Yeah what @drichmond said. And actually this would cause a bug if you need to load multiple programs within one simulation.

mrutt92:
For example, the only reason this test passes is because it happens to load the same binary twice https://github.com/bespoke-silicon-group/bsg_replicant/blob/master/examples/cuda/test_multiple_binary_load.c

drichmond:
@mrutt92 - Wouldn't the icache be re-loaded on the second invocation?

vb000:
Cool, two issues:
- Multiple program inits
- Small program run slower

For the first one, we could find a way to check if are doing cold boot and only use DMA then. Program init during cold boot it the common case so we'd be optimizing the common case.

Second, since we are trying to optimize the architecture & kernels for full workloads or applications, in which there would be about ~100 kernel calls each handling lot of data, we should ideally be measure microbenchmarks without vcache warmup? @drichmond?

mrutt92:
@drichmond yes but with an example in which the code doesn’t fit in icache you’d run into a problem.

drichmond:
You would? I'm not following. The icaches are reloaded on each binary init, the column caches are invalidated. Where do you see the bug?

mrutt92:
When are the column caches invalidated? What code does that?

vb000:
This would cause the bug:
https://github.com/bespoke-silicon-group/bsg_replicant/blob/0d1015b473b7655264b84daf3eb7c7a65123aacf/libraries/bsg_manycore_eva.cpp#L1078-L1079

I understand this implementation would invokes this function and intent was the same too. I didn't consider multiple program inits. @mrutt92 could confirm.

Hence the solution would be find if we are doing a cold boot and DMA only on cold boot.

drichmond:
> When are the column caches invalidated? What code does that?

Doesn't that happen here? https://github.com/bespoke-silicon-group/bsg_replicant/blob/0d1015b473b7655264b84daf3eb7c7a65123aacf/libraries/bsg_manycore.cpp#L1338

vb000:
This implementation won't call `hb_mc_manycore_dma_write`, but it would call this function:
https://github.com/bespoke-silicon-group/bsg_replicant/blob/0d1015b473b7655264b84daf3eb7c7a65123aacf/libraries/bsg_manycore.cpp#L1354

mrutt92:
That's right - `hb_mc_manycore_dma_write` doesn't invalidate cache. If you look at the CUDA functions you'll that they make the invalidate/flush calls.

drichmond:
OK, so we still need to invalidate.

drichmond:
Do we still want to merge this? Or is this a dead end?

PR #630, opened 2020-07-29
Hardware Tile Group Shared Memory

bornaehsani:
- Adding hardware tile group shared memory 
- Regression: hardware tile group shared memory load store
- Regression: hardware tile group shared memory built-in reduction 
- Regression: 1D and 2D versions of blocked matrix multiplication with hardware tile group shared memory
- Host API now also sets two new CSRs for tile group dimensions upon tile group launch.

Cosim regression passes with 4x4_fast_n_fake, 4x4_blocking_vcache_f1_model, and timing_v0_8_4.

**Merge with PR # 370 in bsg_manycore.**


examples/cuda/Makefile
```diff
@@ -135,7 +140,7 @@ include $(EXAMPLES_PATH)/flow.mk
 
 # The default Tile Group size is 2x2
 TILE_GROUP_DIM_X = 2
-TILE_GROUP_DIM_Y = 2
+TILE_GROUP_DIM_Y = 2
```
drichmond, line 143: Extra space -- but doesn't one of your tile group tests use a 4x4?

PR #635, opened 2020-09-02
Remove nonsynthesizable code from aws-fpga top level

drichmond:
Since we've switched over to `dpi-vcs`, I think it would be good to remove all of the non-synthesizable code from `aws-fpga`. This means that `aws-vcs` simulates exactly what `aws-fpga` emulates.

If we don't *need* to maintain it, and `dpi-vcs` is a better/faster platform, I think we should proactively clean out cruft. This also makes it easier to modify the AWS F1 image when we want to do that again in the future.

This includes:
* HBM
* Infinite memory
* Clock Switching

This does not include:
* profilers
* Stats parsers


mrutt92:
I didn't realize that infinite memory and HBM were cruft. Don't we still use those models? Where are those module instantiated now?

drichmond:
This is purely for aws-vcs. We should use them inside of dpi-vcs, which is faster and more up-to-date.

drichmond:
Or dpi-verilator, which also supports those features.

mrutt92:
Got it. So each platform has its own top-level and this one is for aws-vcs which you're saying isn't heavily used and is slow and a bit out of date. 

So if I'm understanding you right this platform is moving in the direction of only having one use case which is simulating are FPGA platform.

Removing cruft makes sense to me in that case and I approve the PR.

mrutt92 (approved):
Nice.

drichmond:
Right. It also means that a full test of all machines is much faster.

PR #636, opened 2020-10-08
Optimized MM host code

drichmond:
Host code for optimized MM Kernel. Ported from PyTorch. Initially written by @yodada and @vb000

### DIFF
--- a/examples/cuda/test_mm_opt.cpp
+++ b/examples/cuda/test_mm_opt.cpp
@@ -26,6 +26,7 @@
 // SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #include "test_mm_opt.hpp"
+#define BLOCK_DIM 8 // this block dim needs to match the same marco in the riscv binary
 #define ALLOC_NAME "default_allocator"
 
 void host_mm_opt(hb_mc_host_tensor_t<float> *result,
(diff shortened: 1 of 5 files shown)


examples/cuda/test_mm_opt.hpp
```diff
@@ -0,0 +1,60 @@
...
+#include <unistd.h>
+#include <string.h>
+#include <time.h>
+#include <stdlib.h>
+#include <random>
+#include <limits>
+
+#include "cuda_tests.h"
+
+typedef struct {
```
save-buffer, line 43: Nit: use the same declaration syntax as for `hb_mc_host_tensor_t` (without the typedef)

examples/cuda/test_mm_opt.cpp
```diff
@@ -0,0 +1,401 @@
...
+
+        std::mt19937 generator;
+        generator.seed(42);
+        std::uniform_real_distribution<float> distribution(-1,1);
+
+        //************************************************************
+        // Allocate memory on the host for mat1, mat2, and result
+        // and initialize the first three with random values.
+        //************************************************************
+        Hmat1.data = new float[Hmat1.N];
```
save-buffer, line 151: Probably should call `delete` on these at the end of the program (or wrap in something like an `std::unique_ptr`)

examples/cuda/test_mm_opt.cpp
```diff
@@ -0,0 +1,401 @@
...
+        //************************************************************
+        // Copy mat1 and mat2, from host onto device
+        //************************************************************
+        void *dst, *src;
+
+        // Copy mat1
+        bsg_pr_info("Copying mat1\n");
+        dst = (void *) ((intptr_t) _mat1);
+        src = (void *) &mat1;
+        rc = hb_mc_device_memcpy (&device, dst, src, sizeof(mat1), HB_MC_MEMCPY_TO_DEVICE);
```
mrutt92, line 198: This is a suggestion more than a change request. CUDA lite has a macro that you can wrap around API calls to do error checking and reporting for you. I think it’s BSG_CUDA_CALL - defined in bsg_manycore_cuda.h. It will reduce the number of LOC and it will also print out the line of code that errored out for you.

drichmond:
Superseded by #765

Issue #640, opened 2020-10-23
verilator platform for manycore simulation


alittlerice:
Hi,

I'm trying to use dpi-verilator for manycore simulation. After setting up as instructed and typing "make regression", I received the following error:

../bsg_bladerunner/bsg_replicant/libraries/bsg_manycore_features.h:38:2: error: #error "_BSG_SOURCE not defined: required for bsg_manycore_runtime"
   38 | #error "_BSG_SOURCE not defined: required for bsg_manycore_runtime"
      |  ^~~~~

Please advise how it could be resolved. Thank you.

drichmond:
Huh, that looks like a typo and should be `_BSD_SOURCE`

What kind of system are you running on?

alittlerice:
Hi Dustin,

It's the correct spelling of the error and it's generated from the file:
bsg_bladerunner/bsg_replicant/libraries/bsg_manycore_features.h

I'm running on ubuntu 20.04. 

Thank you!

alittlerice:
Hi Dustin,

Should we correct the typo in the source file?

I'm running on ubuntu 20.04.

Thank you!

drichmond:
The typo is in our source. It should be _BSD_SOURCE, not _BSG_SOURCE. 

@mrutt92 Can you take a look if you have a moment?

mrutt92:
```c
#ifndef _BSD_SOURCE
#error "_BSG_SOURCE not defined: required for bsg_manycore_runtime"
#endif
```

It's a typo.

mrutt92:
You can correct the typo but it won't make your code compile.  Define _BSD_SOURCE on the command line to fix your problem.

mrutt92:
@alittlerice do you have the full trace from running `make regression`? _BSD_SOURCE is supposed to be defined by the makefile flow but this error would suggest that's not happening.

alittlerice:
@mrutt92  I do have the full trace but it would be too much to attach it here. What info would you want to check and how the _BSD_SOURCE is supposed to be defined? Thank you.

alittlerice:
@mrutt92 ,
Attached is the log file indicating the error. Thank you.

[mylog-.txt](https://github.com/bespoke-silicon-group/bsg_replicant/files/5432389/mylog-.txt)

mrutt92:
The makefile flow is supposed to define it for you.

alittlerice:
[...]

PR #641, opened 2020-10-26
Library Compilation Issues

drichmond:
Addresses #640 

* Support for new GLIBC: Fixed issue with _BSD_SOURCE, now defined in C/CXXFLAGS
* Changed order of linking
* Added std:isnan
* Fixed typo (`_BSG_SOURCE`  -> `_BSD_SOURCE`)


drichmond:
@mrutt92, LGTY?

save-buffer:
One other issue I've run into is the use of `let` in `examples/regression.mk`. Seems that ubuntu uses `dash` as its default shell, and `let` is a `bash` builtin.

drichmond:
Yep - not sure there's something I can do about that right now?

save-buffer:
Instead of `let "total+=1"; \` use `total=$((total+1)); \` according to stack overflow.

drichmond:
Cool. Submit a PR?

save-buffer:
Sure

mrutt92 (approved):
I approve.

PR #642, opened 2020-11-05
Couple of fixups for Ubuntu

save-buffer:
Had to make these changes to make this build in my debian docker container.


examples/python/Makefile
```diff
@@ -66,7 +66,7 @@ DEFINES += -D_XOPEN_SOURCE=500 -D_BSD_SOURCE
 CDEFINES   += $(DEFINES)
 CXXDEFINES += $(DEFINES)
 
-FLAGS     = -g -Wall $(shell python3-config --cflags) -O1 
+FLAGS     = -g -Wall $(shell python3-config --cflags) -O1 -fPIC
```
save-buffer, line 69: I was compiling with Verilator. I'm unsure if we need this for not Verilator, so we may need some conditional logic here.

libraries/libraries.mk
```diff
@@ -102,14 +102,14 @@ $(LIB_OBJECTS): INCLUDES += -I$(LIBRARIES_PATH)/features/dma
 $(LIB_OBJECTS): INCLUDES += -I$(LIBRARIES_PATH)/features/profiler
 # We should move this from AWS (and keep the license)
 $(LIB_OBJECTS): INCLUDES += -I$(AWS_FPGA_REPO_DIR)/SDAccel/userspace/include
-$(LIB_OBJECTS): CFLAGS   := -std=c11 -fPIC -D_GNU_SOURCE $(INCLUDES)
-$(LIB_OBJECTS): CXXFLAGS := -std=c++11 -fPIC -D_GNU_SOURCE $(INCLUDES)
+$(LIB_OBJECTS): CFLAGS    += -std=c11 -fPIC -D_GNU_SOURCE $(INCLUDES) -D_BSD_SOURCE
+$(LIB_OBJECTS): CXXFLAGS  += -std=c++11 -fPIC -D_GNU_SOURCE $(INCLUDES) -D_BSD_SOURCE
```
drichmond, line 106: Pretty sure we don't want += here

libraries/platforms/aws-fpga/hardware/cl_manycore.sv
```diff
@@ -952,9 +952,13 @@ module cl_manycore
...
-    `declare_dramsim3_ch_addr_s_with_pkg(dram_ch_addr_dramsim3_s, `DRAMSIM3_MEM_PKG);
-    dram_ch_addr_dramsim3_s [hbm_num_channels_p-1:0] dramsim3_ch_addr_lo;
-    dram_ch_addr_dramsim3_s [hbm_num_channels_p-1:0] dramsim3_ch_addr_li;
+     // Import the technology-specific address struct from the
+     // appropriate DRAMSIM3 memory package.
+     //
+     // DRAMSIM3_MEM_PKG is defined as a command-line argument to VCS
+     // during the compliation flow for each machine.
+     `DRAMSIM3_MEM_PKG::dram_ch_addr_s [hbm_num_channels_p-1:0] dramsim3_ch_addr_lo;
+     `DRAMSIM3_MEM_PKG::dram_ch_addr_s [hbm_num_channels_p-1:0] dramsim3_ch_addr_li;
```
drichmond, line 961: Changes here would be undone by #635

drichmond:
Can you rebase on master?

libraries/libraries.mk
```diff
@@ -102,14 +102,14 @@ $(LIB_OBJECTS): INCLUDES += -I$(LIBRARIES_PATH)/features/dma
...
 # Need to move this, eventually
 #$(LIB_OBJECTS) $(PLATFORM_OBJECTS): $(BSG_MACHINE_PATH)/bsg_manycore_machine.h
 
 $(LIB_DEBUG_OBJECTS):  CXXFLAGS += -DDEBUG
 
-$(LIB_STRICT_OBJECTS): CXXFLAGS += -Wall -Werror
```
drichmond, line 112: Werror should be kept, unless there's something wrong

environment.mk
```diff
@@ -49,6 +49,7 @@ HARDWARE_PATH    := $(CL_DIR)/hardware
 LIBRARIES_PATH   := $(CL_DIR)/libraries
 MACHINES_PATH    := $(CL_DIR)/machines
 EXAMPLES_PATH    := $(CL_DIR)/examples
+SHELL            := $(shell which bash)
```
drichmond, line 53: This seems reasonable

environment.mk
```diff
@@ -49,6 +49,7 @@ HARDWARE_PATH    := $(CL_DIR)/hardware
...
```
drichmond, line 52: Whoops, this seems reasonable

examples/python/Makefile
drichmond: I don't think it breaks RHEL/Centos, but good to check.

libraries/libraries.mk
save-buffer: Things were exploding, I'll check what specific errors there were, but there were a lot of them so I just got rid of the flag. One possible reason was I believe the default GCC was GCC8 whereas the centos GCC is GCC 4.7 or something like that.

drichmond: I can believe that...

PR #644, opened 2020-11-12
Fixed an issue with using std::isnan, std::isnormal in the header

drichmond:


mrutt92:
When code starts to look like this, it's usually a sign to me that it's time to move to it out of a header and into its own source file. This is fine and it solves your problem for now but it's a symptom of a bigger problem which is this interface needs to be source compatible with both C and C++. This type of work around will start to get old.

drichmond:
Yep. Agree on all points. So much so that I've filed an issue. 

Also, not worth the candle right now.

mrutt92 (approved):
I can respect that.

PR #657, opened 2020-12-29
Boost Verilator

drichmond:
Pulling forward these changes for future reference & possible inclusion. This was motivated by Dan suggesting that I profile the network for our network simulator.

This is a couple months work of on-and-off hacking. I figured out how to profile Verilator (not straightforward...), and then implemented some fixes.

The two significant changes: 

- Replaced bsg_mux_one_hot.v  from basejump with a simulation-amenable version. (FOR VERILATOR ONLY)
- Replaced bsg_encode_one_hot.v  from basejump with a simulation-amenable version. (FOR VERILATOR ONLY)

mux_one_hot was consuming 50% of the execution time for network heavy benchmarks. Encode_one_hot was problematic because it has what Verilator calls an "unopt-flat" issue that can't be "fixed" without substantially re-writing our tape-out RTL. 

I also added -O2 flags for GCC and Verilator. This makes Verilator take a long time to compile (20 minutes) but this process only needs to be done once if someone is not RTL hacking. (otherwise VCS is a better solution)

I left behind some hints for threading and profiling as well.

I haven't done a full analysis, but what I found is that some testbench programs run as fast as VCS now.


taylor-bsg:
Note the behavior is very different if multiple select lines are set. Very
scary. I would suggest an implementation that uses an OR instead of
assignment...

Also a initial begin warning that this is deviating from the synthesized
source base...

M

taylor-bsg:
But very intriguing about the performance differences!

drichmond:
This is

drichmond:
Oops, did not mean to close.

> Note the behavior is very different if multiple select lines are set. Very scary. I would suggest an implementation that uses an OR instead of assignment... Also a initial begin warning that this is deviating from the synthesized source base... M

This is good feedback, I'll implement it after the network simulator.

I also don't know why these modules are _so_ costly. That's something I didn't explore, either.

PR #659, opened 2021-01-08
Use logic & bit wires for nonsynth reset & clk generator

drichmond:
This PR with [#378](https://github.com/bespoke-silicon-group/basejump_stl/pull/378) fixes this repository for new versions of VCS. 

Casting between logic and bit types now produces extra positive/negative transitions at T=0 in simulation. This can cause unwanted and inaccurate assertions.

The PR in basejump provides bit inputs and outputs for the reset generator. This PR switches the "wire" type to bit, to avoid spurious edges between the clock generator and the reset generator. Backward compatible behavior is maintained for other modules by casting the bit clock into a logic clock in the testbench.

Users can now choose between the bit and logic versions of the clock and reset wires.

Also added a comment describing the behavior.

### DIFF
--- a/libraries/platforms/dpi-verilator/hardware/dpi_top.sv
+++ b/libraries/platforms/dpi-verilator/hardware/dpi_top.sv
@@ -58,13 +58,15 @@ module manycore_tb_top
    logic io_reset;
 
    // bsg_nonsynth_clock_gen and bsg_nonsynth_reset_gen BOTH have bit
-   // inputs and outputs (they're non-syntheizable). Casting between
-   // logic and bit can produce unexpected edges at logic types switch
-   // from X to 0/1 at Time 0 in simulation. Therefore, we use bit_clk
-   // and bit_reset for the inputs/outputs of these modules to avoid
-   // unexpected negative/positive edges and other modules can choose
-   // between bit version (for non-synthesizable modules) and the
-   // logic version (otherwise).
+   // inputs and outputs (they're non-synthesizable). Casting between
+   // logic and bit can produce unexpected edges as logic types switch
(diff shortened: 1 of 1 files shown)


libraries/platforms/dpi-verilator/hardware/dpi_top.sv
```diff
@@ -56,7 +56,17 @@ module manycore_tb_top
 
    logic io_clk;
    logic io_reset;
-   
+
+   // bsg_nonsynth_clock_gen and bsg_nonsynth_reset_gen BOTH have bit
+   // inputs and outputs (they're non-syntheizable). Casting between
```
taylor-bsg, line 61: typo

libraries/platforms/dpi-verilator/hardware/dpi_top.sv
```diff
@@ -56,7 +56,17 @@ module manycore_tb_top
...
+   // logic and bit can produce unexpected edges at logic types switch
```
taylor-bsg, line 62: typo

libraries/platforms/dpi-verilator/hardware/dpi_top.sv
drichmond: Done

libraries/platforms/dpi-verilator/hardware/dpi_top.sv
drichmond: Done

drichmond:
Also added a sentence similar to the one in Basejump specifying that the inputs, outputs AND the wire between the module all must be declared bit. Otherwise, casting occurs and we get unexpected neg/pos edges.

taylor-bsg (approved):

PR #660, opened 2021-01-09
Move epa_check_alignment from bsg_manycore.cpp into bsg_manycore_epa.h

drichmond:
Super useful when writing a DPI tile. Also, just generally useful when checking addresses.


mrutt92 (commented):
Not sure why the implementation needs to be in the header.

drichmond:
Doesn't really matter. I figured it was small enough but I'll put it in the .cpp file.

mrutt92 (approved):

PR #663, opened 2021-01-09
Add BSG_MANYCORE_CALL method

drichmond:
CUDA_CALL is nice. I like nice things.

### DIFF
--- a/libraries/bsg_manycore.h
+++ b/libraries/bsg_manycore.h
@@ -171,7 +171,7 @@ extern "C" {
          * @param[in] timeout A timeout counter. Unused - set to -1 to wait forever.
          * @return HB_MC_SUCCESS on success. Otherwise an error code defined in bsg_manycore_errno.h.
          */
-        __attribute__((warn_unused_result, deprecated))
+        __attribute__((warn_unused_result))
         int hb_mc_manycore_packet_rx(hb_mc_manycore_t *mc,
                                      hb_mc_packet_t *packet,
                                      hb_mc_fifo_rx_t type,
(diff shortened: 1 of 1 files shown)


mrutt92 (commented):
Consider making CUDA_CALL invoke this to avoid duplicate code. Where are you planning to use this?

drichmond:
Will do

Just in testing. I got tired of writing if statements that distracted from the purpose of my code.

I was thinking of having it print the name from the manycore struct, too.

mrutt92:
If you do then disregard my previous comment. Library code should print more detailed messages than what's on this macro.

drichmond:
Updated/Addressed

libraries/bsg_manycore.h
```diff
@@ -171,7 +171,7 @@ extern "C" {
          * @param[in] timeout A timeout counter. Unused - set to -1 to wait forever.
          * @return HB_MC_SUCCESS on success. Otherwise an error code defined in bsg_manycore_errno.h.
          */
-        __attribute__((warn_unused_result, deprecated))
```
mrutt92, line 174: just out of curiosity, why was this deprecated in the first place?

mrutt92 (approved):

libraries/bsg_manycore.h
drichmond: To summarize on slack - I think this was deprecated a long time ago. We don't want users to call it, but we still call it in our tests so that we can "finish" the tests.

Our long term goal (as demonstrated in #665) is to replace a call to this with a function for explicit finish/fail.

PR #664, opened 2021-01-12
Callable trigger for SAIF file generation

marcusnchow:
Connected to [pr#454 in bsg_manycore](https://github.com/bespoke-silicon-group/bsg_manycore/pull/454)

Binds saif_generator module to vanilla_core
adds debug statement to support saif generation

PR #666, opened 2021-01-13
Add parameter to disable profilers

drichmond:
Credit @tommydcjung for the seed of this idea. 

This is necessary for two ongoing projects: The network simulator where there _are_ no tiles to bind to, and SAIF generation where the profilers cause VCS to crash.

### DIFF
--- a/machines/baseline_32_16/Makefile.machine.include
+++ b/machines/baseline_32_16/Makefile.machine.include
@@ -55,6 +55,9 @@ BSG_MACHINE_IO_EP_CREDITS             = 16
 BSG_MACHINE_IO_HOST_CREDITS           = $(BSG_MACHINE_IO_EP_CREDITS)
 BSG_MACHINE_IO_REMOTE_LOAD_CAP        = $(BSG_MACHINE_IO_EP_CREDITS)
 
+# (Don't) Instantiate Profilers (needed for SAIF generation, DPI Tile)
+BSG_MACHINE_NOINST_PROFILERS          = 0
+
 ##################### Constants and Computations #####################
 # Cannot be changed for xbar networks:
 BSG_MACHINE_HETERO_TYPE_VEC           = default:0
--- a/hardware/hardware.mk
+++ b/hardware/hardware.mk
@@ -280,6 +280,7 @@ $(BSG_MACHINE_PATH)/bsg_bladerunner_pkg.v: $(BSG_MACHINE_PATH)/bsg_bladerunner_c
 	@echo "parameter bsg_machine_noc_y_max_gp = bsg_machine_core_y_global_max_gp + 2; // Caches on both sides" >> $@
 	@echo "parameter bsg_machine_noc_x_coord_width_gp = \`BSG_SAFE_CLOG2(bsg_machine_noc_x_max_gp);" >> $@
 	@echo "parameter bsg_machine_noc_y_coord_width_gp = \`BSG_SAFE_CLOG2(bsg_machine_noc_y_max_gp);" >> $@
+	@echo "parameter bsg_machine_noinst_profilers_gp = $(BSG_MACHINE_NOINST_PROFILERS);" >> $@
 	@echo >> $@
 	@echo "endpackage" >> $@
 	@echo >> $@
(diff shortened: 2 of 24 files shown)


marcusnchow:
with regards to [#664](https://github.com/bespoke-silicon-group/bsg_replicant/pull/664)  can we add a warning somewhere to state that the profiler and saif generation can't be used at the same time. I can also modify my changes to use the new parameter to make sure that saif is disable while profiling

drichmond:
yeah, one thing we may want to do is have a parameter that enables saif generation. If both parameters are set, multiple errors/warnings would occur.

drichmond:
We don't want to run VCS with `-debug_pp`. It's pretty hefty hit for simulation speed.

machines/baseline_32_16/Makefile.machine.include
```diff
@@ -55,6 +55,9 @@ BSG_MACHINE_IO_EP_CREDITS             = 16
 BSG_MACHINE_IO_HOST_CREDITS           = $(BSG_MACHINE_IO_EP_CREDITS)
 BSG_MACHINE_IO_REMOTE_LOAD_CAP        = $(BSG_MACHINE_IO_EP_CREDITS)
 
+# (Don't) Instantiate Profilers (needed for SAIF generation, DPI Tile)
+BSG_MACHINE_NOINST_PROFILERS          = 0
```
tommydcjung, line 59: I don't think this should be a machine parameter, because it's not a parameter that changes hw. It should instead be a flag when building simv.

drichmond: Why not?

mrutt92: I agree with @tommydcjung - where do we normally turn on profiling flags? I don't recall it being in the machine makefile.

drichmond: @mrutt92 Profiling, or trace?

drichmond: We turn on/off the tracers using a C API now, instead of using +trace.

That doesn't prevent the tracer from being instantiated. We need a solution that prevents the tracer from being instantiated.

mrutt92: `ifdef ?

drichmond: Not instantiating the profilers also affects the runtime library. The library expects DPI calls to be available that return the number of FLOPS, IOPS, and cycles. If the profilers are not instantiated then the linker will fail.

mrutt92: __attribute__((weak)) ?

drichmond: `ifdef would work to remove them. We could do a debug/saif solution like I mentioned above -- but we still need a solution for the C library.

mrutt92:
```c
__attribute__((weak))
int whatever_function_would_be_missing_if_ifdef_the_profilers_out()
{
    return NOIMPL;
}
```

drichmond: Hmmm. What if you compile both the .saif and normal version? Is there a chance the normal version can still use the weak method?

drichmond: Alternatively, we can put a macro-guarded definition in the profiler header.

mrutt92: Sure. But now that I'm thinking about this, here's an argument for keeping the flag in the Makefile.machine.include instead of some random option: People know how to find/use the Makefile.machine.include.

taylor-bsg:
Just to summarize, we discussed, it is better to keep it out of the
Makefiles since it is not machine configuration state and will lead to a
proliferation of files. However we should generate separate binaries (or at
least ones with different names) to avoid accidentally forgetting to blow
away a stale binary with a non current set of params baked in.

drichmond:
yep. Good to record our consensus.

drichmond:
Replaced by #694

PR #670, opened 2021-01-21
WIP: Bigblade VCS Simulation

drichmond:
New simulation files for BigBlade

This is WIP at the moment. 

What's working

- Compilation (except for DPI interface)

What's not working: 

- The DPI Interface....

The issue is the endpoint to fifo interface. The packet format has changed and this module needs to be updated. I'll get to it tomorrow and clean this PR up a bit.


drichmond:
OK, this PR is now compiling. it doesn't pass test_rom yet, because it doesn't recognize the memory configuration.

Other known issues: 

1. The ROM needs to have the pod dimensions added to it.
2. The packet format has changed and the C library will need to be fixed.
3. The RTL hierarchy has changed and the DPI library will need to be fixed.

There may be issues beyond these. 

I will work on addressing the memory configuration and 3, first, since they need to be fixed before solving 1 and 2.

drichmond:
test_rom passes, with timing_mesh_8_4, but:

- The memory system of infinte_mesh_8_4 (e_vcache_test_mem) is not recognized by the C++ library
- The origin coordinate is incorrect (this needs to be fixed in the machine file)
- The host coordinate is incorrect
- The Host C++ library is sending DRAM enable packets before reset is complete -- and causing assertions. (bsg_manycore_link_to_fifos.v) -- commented out to get to pass.
- The DPI interface needs a reset_done signal

examples/library/test_get_cycle.cpp
```diff
@@ -55,12 +55,8 @@ int test_get_cycle (int argc, char **argv) {
...
 
         // To cause time to proceed, we write to DRAM.
-        uint32_t manycore_dim_x = hb_mc_coordinate_get_x(hb_mc_config_get_dimension_vcore(config));
-        uint32_t manycore_dim_y = hb_mc_coordinate_get_y(hb_mc_config_get_dimension_vcore(config));
-
-        uint32_t dram_coord_x = 0;
-        uint32_t dram_coord_y = hb_mc_config_get_dram_y(config);
-        hb_mc_npa_t npa = { .x = dram_coord_x, .y = dram_coord_y, .epa = 0};
+        hb_mc_coordinate_t dram_coord = hb_mc_config_get_dram_coordinate(config, 0);
+        hb_mc_npa_t npa = hb_mc_npa(dram_coord, 0);
```
drichmond, line 59: Good. I think we have an EPA DRAM, too. I thought I added it but it might be in my network simulator PR

examples/library/test_manycore_dmem_read_write.cpp
```diff
@@ -56,6 +56,7 @@ int test_mem_functions(hb_mc_manycore_t *mc) {
         }
         bsg_pr_test_info("Write successful\n");
 
+        // TODO: We need to move available IDs to global.
```
drichmond, line 59: Did I write this?

mrutt92: yes

examples/library/test_manycore_dmem_read_write.cpp
```diff
@@ -165,6 +165,7 @@ int test_manycore_dmem_read_write () {
         }
 
         int r = HB_MC_FAIL;
+
```
drichmond, line 168: TODO: Fix added spaces

examples/library/test_manycore_dmem_read_write.cpp
drichmond: OK, less of a comment for this test, and more of a comment for the library. TODO: Remove this comment and file an issue to track outstanding IDs at the pod (?) or device (?) level

drichmond: I do this in the network simulator with a set, but there are many ways to do this.

examples/library/test_get_cycle.cpp
drichmond: https://github.com/bespoke-silicon-group/bsg_replicant/blob/2af7bfecabacd28349ec3362dbb387a8d39dab78/libraries/bsg_manycore_vcache.h#L50

examples/library/test_manycore_parameters.h
```diff
@@ -49,12 +49,12 @@
 #define GLOBAL_INDICATOR_WIDTH 1
 #define GLOBAL_INDICATOR (1 << 30)
 #define GLOBAL_EPA_SIZE DMEM_EPA_SIZE
-#define GLOBAL_X_BITS 6
+#define GLOBAL_X_BITS 7
```
drichmond, line 52: Do we still run this test? Is this the one you hated with a passion?

examples/library/test_manycore_vcache_sequence.c
```diff
@@ -56,16 +56,17 @@ int test_manycore_vcache_sequence() {
...
-        uint32_t dram_coord_y = hb_mc_config_get_dram_y(config);
         int mismatch = 0;
 
         /**************************************************************/
         /* Loop over all DRAM banks and write ARRAY_LEN words to each */
         /**************************************************************/
-        for (dram_coord_x = 0; dram_coord_x < 1; dram_coord_x++) {
+        hb_mc_coordinate_t dram_coord;
+        hb_mc_config_foreach_dram_coordinate(dram_coord, config)
+        {
```
drichmond, line 66: This is pretty excellent.

examples/library/test_manycore_parameters.h
mrutt92: Yes. I do hate this test because all it really tests is that we updated the macros here.

examples/library/test_manycore_vcache_sequence.c
mrutt92: Yeah... I remembered having to update these tests when we added vcaches to the north of the array. I figured we could do this and it might spare us the pain going forward. There is some code bsg_manycore.cpp that could potentially make use of these, too.

mrutt92: Macros like this are pretty common in C code because they hide loop logic that distracts from the logic of the program. It should be noted that modern C++ does the same thing (and clearly better IMO) with this syntax:
```c++
for (type iter : container) { }
```

mrutt92: OTHO the lines of code required to write a complex random access iterator might be more verbose than this macro so...

examples/library/test_manycore_dmem_read_write.cpp
mrutt92: I think we do it now with an std::map (this PR) but it's lifetime is at the function scope. CUDA-lite does not support a non-blocking read api. If that were something we wanted to support, then persistent tracking of load ids would be necessary. OTOH that API functionality comes with additional headaches. For one, non-blocking APIs are much harder to use.

examples/library/test_manycore_vcache_sequence.c
drichmond: Do what you feel is best

examples/library/test_manycore_dmem_read_write.cpp
drichmond: Yeah - I think I'm more concerned about a user manually sending packets, and then using the read/write API. 

I still feel that we should file an issue and track it. Would be an easy fix.

examples/library/test_packet.h
```diff
@@ -0,0 +1,32 @@
...
+// (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+// LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+// ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+// SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#pragma once
+#include <bsg_manycore.h>
+#include <inttypes.h>
+#include "../cl_manycore_regression.h"
```
drichmond, line 31: I do plan to get rid of this .h requirement... someday

examples/library/test_vcache_sequence.c
```diff
@@ -44,18 +44,19 @@ int test_vcache_sequence() {
...
-
+        int dram_count = 0, dram_break_after = 1;
+        hb_mc_coordinate_t dram;
+        hb_mc_config_foreach_dram_coordinate(dram, config)
+        {
+                if (dram_count++ >= dram_break_after)
+                        break;
+
+                /* To check all dram banks change 1 to manycore_dim_x */
+                uint32_t dram_coord_x, dram_coord_y;
```
drichmond, line 57: Why wouldn't we want to? Takes too long?

mrutt92: I guess so. I didn't write the original test, I just made it pass.

examples/library/test_manycore_parameters.h
drichmond: I'm fine with nuking this test

examples/library/test_manycore_vcache_sequence.c
mrutt92: Well there's no choice here: We're a C API.

libraries/bsg_manycore_config.cpp
```diff
@@ -130,21 +130,28 @@ int hb_mc_config_init(const hb_mc_config_raw_t raw[HB_MC_CONFIG_MAX],
...
-        if ((idx < HB_MC_COORDINATE_MIN) || (idx > config->vcore_dimensions.x)){
-                bsg_pr_err("%s: Invalid Host Interface index X %d: %s\n",
-                           __func__, idx, error_init_help);
-                return HB_MC_INVALID;
-        }
+        // if ((idx < HB_MC_COORDINATE_MIN) || (idx > config->vcore_dimensions.x)){
+        //         bsg_pr_err("%s: Invalid Host Interface index X %d: %s\n",
+        //                    __func__, idx, error_init_help);
+        //         return HB_MC_INVALID;
+        // }
```
drichmond, line 137: Comments? Remove?

examples/library/test_vcache_sequence.c
mrutt92: admittedly the comment should be axed.

libraries/bsg_manycore_config.cpp
```diff
@@ -130,21 +130,28 @@ int hb_mc_config_init(const hb_mc_config_raw_t raw[HB_MC_CONFIG_MAX],
...
         config->host_interface.y = idx;
 
+
+        idx = raw[HB_MC_CONFIG_NOC_COORD_X_WIDTH];
+        config->noc_coord_width.x = idx;
+
+        idx = raw[HB_MC_CONFIG_NOC_COORD_Y_WIDTH];
+        config->noc_coord_width.y = idx;
+
```
drichmond, line 154: Check validity?

libraries/bsg_manycore_config.h
```diff
@@ -58,8 +58,8 @@ extern "C" {
         #define HB_MC_CONFIG_MAX_BITWIDTH_ADDR 30
         #define HB_MC_CONFIG_MAX_BITWIDTH_DATA 32
 
-        #define HB_MC_CONFIG_VCORE_BASE_X 0
-        #define HB_MC_CONFIG_VCORE_BASE_Y 2
+        #define HB_MC_CONFIG_VCORE_BASE_X 16
+        #define HB_MC_CONFIG_VCORE_BASE_Y 8
```
drichmond, line 62: is this read from the rom now?

libraries/bsg_manycore_config.cpp
mrutt92: Tell me what "validity" means and I'll add the check.

libraries/bsg_manycore_config.h
```diff
@@ -238,11 +218,14 @@ extern "C" {
...
+        static inline hb_mc_dimension_t hb_mc_config_get_noc_coord_width(const hb_mc_config_t *cfg){
+                return cfg->noc_coord_width;
+        }
+
         static inline hb_mc_dimension_t hb_mc_config_get_dimension_network(const hb_mc_config_t *cfg){
-                hb_mc_dimension_t dim = hb_mc_config_get_dimension_vcore(cfg);
-                // The Network has three additional Y rows: An IO Row, and two DRAM/Cache Rows
-                return hb_mc_dimension(hb_mc_dimension_get_x(dim),
-                                       hb_mc_dimension_get_y(dim) + 3);
+                hb_mc_dimension_t noc_width = hb_mc_config_get_noc_coord_width(cfg);
```
drichmond, line 226: dimension_t for noc width?

libraries/bsg_manycore_config.h
```diff
@@ -320,13 +300,19 @@ extern "C" {
         static inline hb_mc_coordinate_t
         hb_mc_config_get_dram_coordinate(const hb_mc_config_t *cfg, hb_mc_idx_t cache_id)
         {
-                hb_mc_coordinate_t dims;
-                dims = hb_mc_config_get_dimension_network(cfg);
+                hb_mc_coordinate_t dims = hb_mc_config_get_dimension_vcore(cfg);
+                hb_mc_coordinate_t origin = hb_mc_config_get_origin_vcore(cfg);
+
+                hb_mc_idx_t x = cache_id % hb_mc_dimension_get_x(dims) + hb_mc_coordinate_get_x(origin);
+                hb_mc_idx_t y;
```
drichmond, line 307: Parenthesis to make OOO clear

libraries/bsg_manycore_config.cpp
mrutt92: If it means a hardcoded number, we'll just be right back here next tape out. We'll also need separate versions of the library for each chip.

libraries/bsg_manycore_eva.cpp
```diff
@@ -50,18 +50,18 @@
 #define DEFAULT_GROUP_BITIDX (DEFAULT_GROUP_Y_BITIDX + DEFAULT_GROUP_Y_LOGSZ)
 #define DEFAULT_GROUP_BITMASK (1ULL << DEFAULT_GROUP_BITIDX)
 
-#define DEFAULT_GLOBAL_X_LOGSZ 6
-#define DEFAULT_GLOBAL_X_BITIDX HB_MC_EPA_LOGSZ
+#define DEFAULT_GLOBAL_X_LOGSZ 7
```
drichmond, line 53: Can't this be replaced by rom value?

libraries/bsg_manycore_config.h
mrutt92: I guess it can be. I kicked this can down the road because having multiple manycore pods will change what this even means.

libraries/bsg_manycore_config.cpp
drichmond: I'm fine with just checking nonzero and not all 1's. And maybe less than say 32. All of those values are wild error conditions

libraries/bsg_manycore_config.h
mrutt92: not sure what the problem is?

libraries/bsg_manycore_config.h
drichmond: OK, push to #667

libraries/bsg_manycore_eva.cpp
mrutt92: All of this garbage could probably be replaced with ROM values. Is that a good use of our time right now?

libraries/bsg_manycore_config.cpp
mrutt92: at that point why bother checking at all?

examples/library/test_manycore_dram_read_write.c
```diff
@@ -76,15 +81,16 @@ int test_manycore_dram_read_write() {
                         if (i % 64 == 1)
                                 bsg_pr_test_info("%s: Have written %zu words to DRAM\n",
                                     __func__, i);
-                        
-                        hb_mc_npa_t npa = { .x = dram_coord_x, .y = dram_coord_y, .epa = BASE_ADDR + (i*4) };
+
+                        hb_mc_npa_t npa = hb_mc_npa(dram_coord, BASE_ADDR + (i*4));
```
drichmond, line 85: Switch to using HB_MC_VCACHE_EPA_OFFSET_DRAM

examples/library/test_packet.c
```diff
@@ -0,0 +1,98 @@
...
+// ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+// SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#include "bsg_manycore_printing.h"
+#include "test_packet.h"
+#include "bsg_manycore_cuda.h"
+#include <inttypes.h>
+
+#define ADDR (0x1000 >> 2)
```
drichmond, line 33: What is ADDR here?

hardware/memsys_test_mem.mk
```diff
@@ -0,0 +1,42 @@
+# Copyright (c) 2019, University of Washington All rights reserved.
```
drichmond, line 1: Since these .mk files are breeding we should put them somewhere

libraries/bsg_manycore.cpp
```diff
@@ -708,24 +714,15 @@ static int hb_mc_manycore_send_read_rqst(hb_mc_manycore_t *mc,
...
-                                                      HB_MC_PACKET_REQUEST_MASK_BYTE << shift));
-                break;
-        default:
-                return HB_MC_INVALID;
-        }
+
+        // set load info
+        hb_mc_request_packet_load_info_t info = {};
+        info.part_sel       = hb_mc_npa_get_epa(npa) & 0x3;
+        info.is_unsigned_op = 1;
```
drichmond, line 721: Hmmmm. I think we did this before, but don't we have a templated read function somewhere?

drichmond: Plus a test to verify these work

libraries/bsg_manycore_config.cpp
drichmond: A lot of these are smoke tests - like if we mapped an AXI platform incorrectly or the ROM is the wrong size.

libraries/bsg_manycore_config.h
drichmond: Seems odd to use the dimension type for something that isn't is a dimension

libraries/bsg_manycore_eva.cpp
drichmond: This one in particular is in the rom, no? The x/y coordinate width?

libraries/bsg_manycore_memsys.c
```diff
@@ -16,6 +16,8 @@ const char * hb_mc_memsys_id_to_string(hb_mc_memsys_id_t id)
                 [HB_MC_MEMSYS_ID_AXI4] = "AXI4 Memory Controller",
                 [HB_MC_MEMSYS_ID_INFMEM] = "Ideal Memory System",
                 [HB_MC_MEMSYS_ID_DRAMSIM3] = "DRAMSim3 HBM2",
+                [HB_MC_MEMSYS_ID_TESTMEM] = "Non-synthesizable Test Memory",
+                [HB_MC_MEMSYS_ID_HBM2] = "High Bandwidth Memory 2",
```
drichmond, line 20: I would add "(via Wormhole Network)" to make it clear how these are different

libraries/bsg_manycore_request_packet.h
```diff
@@ -58,19 +58,65 @@ extern "C" {
...
+
+#define HB_MC_REQUEST_PACKET_LOAD_INFO_GET(data, field)                 \
+                (((data) & HB_MC_REQUEST_PACKET_LOAD_INFO_ ## field ##_MASK) >> \
+                 HB_MC_REQUEST_PACKET_LOAD_INFO_ ## field ## _SHIFT)
+
+#define HB_MC_REQUEST_PACKET_LOAD_INFO_SET(data, field, field_val)      \
+        do {                                                            \
+                data &= ~HB_MC_REQUEST_PACKET_LOAD_INFO_ ## field ## _MASK; \
+                data |= ((field_val) << HB_MC_REQUEST_PACKET_LOAD_INFO_ ## field ## _SHIFT) & \
+                        HB_MC_REQUEST_PACKET_LOAD_INFO_ ## field ## _MASK; \
```
drichmond, line 103: Is this what you meant when you said you went whole hog on macros?

examples/library/test_packet.c
mrutt92: Word EPA

libraries/libraries.mk
```diff
@@ -86,7 +87,11 @@ LIB_STRICT_OBJECTS += $(LIBRARIES_PATH)/bsg_manycore_print_int_responder.o
 LIB_STRICT_OBJECTS += $(LIBRARIES_PATH)/bsg_manycore_memsys.o
 
 # Objects that should be compiled with debug flags
-LIB_DEBUG_OBJECTS  +=
+# LIB_DEBUG_OBJECTS  += $(LIBRARIES_PATH)/bsg_manycore_config.o
+# LIB_DEBUG_OBJECTS  += $(LIBRARIES_PATH)/bsg_manycore_memsys.o
+# LIB_DEBUG_OBJECTS  += $(LIBRARIES_PATH)/bsg_manycore.o
+# LIB_DEBUG_OBJECTS  += $(LIBRARIES_PATH)/bsg_manycore_eva.o
+# LIB_DEBUG_OBJECTS  += $(LIBRARIES_PATH)/bsg_manycore_cuda.o
```
drichmond, line 94: Remove

libraries/platforms/aws-fpga/hardware/bsg_manycore_endpoint_to_fifos.v
```diff
@@ -35,18 +35,14 @@
 
 module bsg_manycore_endpoint_to_fifos
   import bsg_manycore_pkg::*;
-  import bsg_manycore_link_to_axil_pkg::*;
+//  import bsg_manycore_link_to_axil_pkg::*;
```
drichmond, line 38: Comment

hardware/memsys_test_mem.mk
mrutt92: You mean we should take the old ones out to the farm?

libraries/platforms/aws-fpga/hardware/bsg_manycore_endpoint_to_fifos.v
```diff
@@ -128,33 +130,60 @@ module bsg_manycore_endpoint_to_fifos
...
+  // end
+
+  // always @(posedge clk_i) begin
+  //   if (mc_rsp_v_o & mc_rsp_ready_i) begin
+  //     $display("bsg_manycore_endpoint_to_fifos (response): type=%s", returned_pkt_type_r_lo.name());
+  //     $display("bsg_manycore_endpoint_to_fifos (response): data=%h", mc_rsp_lo_cast.data);
+  //     $display("bsg_manycore_endpoint_to_fifos (response): reg_id=%h", mc_rsp_lo_cast.reg_id);
+  //   end
+  // end
+
```
drichmond, line 180: debug_p?

libraries/bsg_manycore.cpp
mrutt92: It calls this.

libraries/platforms/bigblade-vcs/hardware/dpi_top.sv
```diff
@@ -0,0 +1,308 @@
...
+       )
+   print_stat_snoop
+     (
+      .loader_link_sif_in_i(host_link_sif_lo)
+      ,.loader_link_sif_out_i(host_link_sif_li)
+
+      ,.print_stat_v_o(print_stat_v)
+      ,.print_stat_tag_o(print_stat_tag)
+      );
+*/
```
drichmond, line 283: I think this needs to be uncommented

libraries/platforms/bigblade-vcs/hardware/dpi_top.sv
```diff
@@ -0,0 +1,308 @@
...
+       ,.rom_els_p(bsg_machine_rom_els_gp)
+       ,.rom_width_p(bsg_machine_rom_width_gp)
+       ,.rom_arr_p(bsg_machine_rom_arr_gp)
+       ,.max_out_credits_p(bsg_machine_io_credits_max_gp)
+       )
+   mc_dpi
+     (
+      .clk_i(host_clk)
+      // DR: I don't particularly like this, but we'll leave it for now
+      ,.reset_i(host_reset | (~core_reset_done_r))
```
drichmond, line 237: I don't like this - but I don't have a better solution for now. Open to suggestions

Not including core_reset_done means that the endpoint in the host interface will assert when it comes out of reset before the manycore. Perhaps the best solution is to punt this logic internally

libraries/platforms/dpi-verilator/bsg_manycore_platform.cpp
```diff
@@ -545,3 +547,19 @@ int hb_mc_platform_log_disable(hb_mc_manycore_t *mc){
...
+ * Check if chip reset has completed.
+ * @param[in] mc    A manycore instance initialized with hb_mc_manycore_init()
+ * @return HB_MC_SUCCESS on success. Otherwise an error code defined in bsg_manycore_errno.h.
+ */        
+int hb_mc_platform_reset_is_done(hb_mc_manycore_t *mc)
+{
+        hb_mc_platform_t *pl = reinterpret_cast<hb_mc_platform_t*>(mc->platform);
+        manycore_pr_dbg(mc, "%s: calling eval()\n", __func__);
+        pl->top->eval();
+        bool done;
```
drichmond, line 561: I'm not sure this is the right place for this eval call. Did you remove the reset done check from tx? There was a while loop in init, right?

machines/archive/infinite_mesh_8_4/Makefile.machine.include
```diff
@@ -27,6 +27,10 @@
 
 # Change these parameters to define your machine. All other parameters should remain constant.
 
+#
+BSG_MACHINE_NUM_PODS_X                = 1
+BSG_MACHINE_NUM_PODS_Y                = 1
+
```
drichmond, line 33: We should undo the changes to this file

machines/archive/timing_mesh_8_4/Makefile.machine.include
```diff
@@ -36,9 +36,9 @@
 
 # Change these parameters to define your machine. All other parameters should remain constant.
 
-# Manycore X/Y dimensions
-BSG_MACHINE_NUM_CORES_X               = 8
-BSG_MACHINE_NUM_CORES_Y               = 4
+# Chip Pod Dimensions
+BSG_MACHINE_DIM_PODS_X                = 1
+BSG_MACHINE_DIM_PODS_Y                = 1
```
drichmond, line 41: And this one

examples/library/test_packet.c
drichmond: For DMEM?

hardware/memsys_test_mem.mk
drichmond: Uh, sure. 

But what I really meant was to put them all in a folder. Least of my concerns.

libraries/bsg_manycore.cpp
drichmond: Then we should have some way of setting is_unsigned. Or file an issue for it.

machines/pod_X1Y1_ruche_X16Y8_hbm/Makefile.machine.include
```diff
@@ -0,0 +1,125 @@
...
+##################### Constants and Computations #####################
+
+# Aliases required for memsys.mk, and others.
+# Specific to the HBM2E chip we're modeling (8GB, 4-stack/8-channel)
+__BSG_MACHINE_DRAMSIM3_CHIP_SIZE_IN_WORDS := $(shell echo 2^31 | bc) # 4GB
+__BSG_MACHINE_DRAMSIM3_CHIP_CH_SIZE_IN_WORDS := $(shell echo $(__BSG_MACHINE_DRAMSIM3_CHIP_SIZE_IN_WORDS)/8 | bc)
+
+# TODO: Specific to pod (next 4)
+BSG_MACHINE_NUM_VCACHE               := $(shell echo $(BSG_MACHINE_POD_TILES_X)*2 | bc)
+BSG_MACHINE_DRAM_CHANNELS            := $(shell echo $(BSG_MACHINE_NUM_VCACHE)/$(BSG_MACHINE_VCACHE_PER_DRAM_CHANNEL) | bc)
```
drichmond, line 102: Need to fix

libraries/bsg_manycore_config.h
drichmond: Oh, I see what you're doing. OK. Fine.

libraries/platforms/dpi-verilator/bsg_manycore_platform.cpp
drichmond: I would rather call this reset_is_done, and make a reset_done_wait function.

examples/library/test_packet.c
mrutt92: Yeah it's a v-core's dmem.

libraries/bsg_manycore.cpp
mrutt92: When do we do a signed read? is_unsigned is not is_word. It's a flag to indicate if the returned data should be sign extended.

mrutt92: AFAIK our model in CUDA-Lite is to read/write data 'as is' and support for sign extension in our API would be a new feature.

examples/library/test_packet.c
drichmond: HB_MC_TILE_EPA_DMEM_BASE is what you're looking for

libraries/bsg_manycore_request_packet.h
mrutt92: Less gross than spaghetti shift and anding.

libraries/bsg_manycore.cpp
drichmond: Worth filing an issue - no urgency to fix

libraries/bsg_manycore_config.cpp
```diff
@@ -130,21 +139,20 @@ int hb_mc_config_init(const hb_mc_config_raw_t raw[HB_MC_CONFIG_MAX],
...
-        if ((idx < HB_MC_COORDINATE_MIN) || (idx > config->vcore_dimensions.y)){
-                bsg_pr_err("%s: Invalid Host Interface index Y %d: %s\n",
-                           __func__, idx, error_init_help);
-                return HB_MC_INVALID;
-        }
         config->host_interface.y = idx;
 
+
+        idx = raw[HB_MC_CONFIG_NOC_COORD_X_WIDTH];
+        CHECK_FIELD(HB_MC_CONFIG_NOC_COORD_X_WIDTH, idx > 0 && idx < 32);
```
drichmond, line 149: Excellent

drichmond: Can you apply it throughout the file?

libraries/bsg_manycore.cpp
mrutt92: I don't even think its worth that. This would be a feature nobody's asking for.

libraries/platforms/bigblade-vcs/hardware/dpi_top.sv
mrutt92: I think you should fix it.

libraries/bsg_manycore.cpp
mrutt92: See test_vcache_flush for a test that exercises this change.

libraries/bsg_manycore_eva.cpp
mrutt92: Kinda. There's also a group width which is inexplicably only 6. The relation between this value and the x-coord width does not feel like 'solid ground' to me.

libraries/platforms/bigblade-vcs/hardware/dpi_top.sv
mrutt92: Please fix or file an issue.

libraries/bsg_manycore.cpp
drichmond: Wait, did you implement it?

libraries/platforms/dpi-verilator/bsg_manycore_platform.cpp
mrutt92: I did remove the reset done check from tx. It's now in init.

libraries/platforms/bigblade-vcs/hardware/dpi_top.sv
drichmond: Yes, and I forgot to when doing my fixes.

examples/library/test_manycore_parameters.h
mrutt92: Let's file an issue; the problem with this test is it duplicates code for materializing group/global EVAs. There's probably a better way to test this.

machines/pod_X1Y1_ruche_X16Y8_hbm/Makefile.machine.include
mrutt92: Go ahead and fix if you know what to do.

libraries/bsg_manycore_eva.cpp
drichmond: That suggests there should be a group width field in the rom... is the group with just implicit?

libraries/platforms/bigblade-vcs/hardware/dpi_top.sv
drichmond: Done

machines/pod_X1Y1_ruche_X16Y8_hbm/Makefile.machine.include
drichmond: We need to enumerate the number of vcaches. This is only partially correct. Maybe it's as simple as multiplying by pod Y and X

libraries/platforms/bigblade-vcs/hardware/dpi_top.sv
drichmond: Done, #678

machines/pod_X1Y1_ruche_X16Y8_hbm/Makefile.machine.include
drichmond: Fixed

PR #693, opened 2021-03-09
Reworked makefiles to facilitate per-test folder/Makefile

drichmond:
* Removed dependency on .h/.hpp for compilation
* Changed simulator binary target to .profile


drichmond:
The did yesterday, i'll run it again for luck

drichmond:
Passes

PR #696, opened 2021-03-11
Slightly refactor store request packet call-tree

drichmond:
Packets should use SW, but this required passing sz to the function that set the op. This seemed to break an abstraction so I merged hb_mc_manycore_format_store_request_packet into its single callsite and set SW/STORE there.

Also, refactored to use HB_MC_SUCCESS instead of 0.

### DIFF
--- a/libraries/bsg_manycore.cpp
+++ b/libraries/bsg_manycore.cpp
@@ -422,21 +422,7 @@ static int hb_mc_manycore_format_request_packet(hb_mc_manycore_t *mc, hb_mc_requ
         hb_mc_request_packet_set_x_src(pkt, hb_mc_coordinate_get_x(host_coordinate));
         hb_mc_request_packet_set_y_src(pkt, hb_mc_coordinate_get_y(host_coordinate));
         hb_mc_request_packet_set_addr(pkt, hb_mc_npa_get_epa(npa) >> 2);
-        return 0;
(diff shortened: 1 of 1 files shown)


drichmond:
(I have not run regression on this yet, but will post when I have)

libraries/bsg_manycore.cpp
```diff
@@ -445,13 +431,13 @@ static int hb_mc_manycore_format_load_request_packet(hb_mc_manycore_t *mc,
 {
         int r;
 
-        if ((r = hb_mc_manycore_format_request_packet(mc, pkt, npa)) != 0)
+        if ((r = hb_mc_manycore_format_request_packet(mc, pkt, npa)) HB_MC_SUCCESS)
```
dpetrisko, line 434: dropped an != ?

libraries/bsg_manycore.cpp
```diff
@@ -461,13 +447,13 @@ static int hb_mc_manycore_format_cache_op_request_packet(hb_mc_manycore_t *mc,
...
```
dpetrisko, line 450: !=

drichmond: HA. Shows I didn't even run compilation. Yes, I screwed something up in my rebase.

PR #703, opened 2021-03-17
Fixed a bug in EVA to NPA translation code

drichmond:
* dram_x_min/max assumed that an origin was passed.

### DIFF
--- a/libraries/bsg_manycore_eva.cpp
+++ b/libraries/bsg_manycore_eva.cpp
@@ -335,15 +335,19 @@ static bool default_eva_is_dram(const hb_mc_eva_t *eva)
         return (hb_mc_eva_addr(eva) & DEFAULT_DRAM_BITMASK) != 0;
 }
 
-static uint32_t default_dram_max_x_coord(const hb_mc_config_t *cfg, const hb_mc_coordinate_t *origin)
+static uint32_t default_dram_max_x_coord(const hb_mc_config_t *cfg, const hb_mc_coordinate_t *tgt)
 {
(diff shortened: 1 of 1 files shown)


libraries/bsg_manycore_eva.cpp
```diff
@@ -396,28 +400,31 @@ static int default_eva_get_x_coord_dram(const hb_mc_manycore_t *mc,
                                         const hb_mc_eva_t *eva,
                                         hb_mc_idx_t *x) {
         hb_mc_coordinate_t pod = hb_mc_config_pod(cfg, *src);
-#ifdef DEBUG
+        hb_mc_coordinate_t og = hb_mc_config_pod_vcore_origin(cfg, pod);
+#ifndef DEBUG
```
mrutt92, line 404: fix

examples/library/test_printing/main.c
```diff
@@ -50,6 +50,7 @@
 #include <time.h>
 #include <stdlib.h>
 #include <bsg_manycore.h>
+#include <bsg_manycore_printing.h>
```
mrutt92, line 53: why is this here?

libraries/bsg_manycore_eva.cpp
drichmond: Oops, thought I undid this. Fixed

examples/library/test_printing/main.c
drichmond: Fails silently with a linker error if it isn't.

drichmond: I'll pull it out into a separate PR

mrutt92 (approved):

PR #488, opened 2020-01-17
Added test_vcache_atomic_histogram to cuda tests

drichmond:

PR #709, opened 2021-03-26
Support for memory system refactor in bsg_manycore

drichmond:
Partially Working Support for memory system refactor in bsg_manycore (#488)

* Added new machine parameters
* Added new DRAMSIM3 Package macro
* Added clean targets

DMA Tests currently fail


drichmond:
Update: test_high_mem and test_dma currently fail. This feels like a south-side cache-DMA mapping issue

drichmond:
Passes regression. 

This supports the new RTL, but does not enable multiple rows of caches for a pod. This is PR 1 of 2.

PR #711, opened 2021-04-01
Removing cruft headers from library tests and guarding mutexes in the memory manager

sripathi-muralitharan:
This PR removes cruft headers from the library tests and adds a macro guard around the use of mutexes. 

- Removes the use of sys/ioctl.h
- Removes the use of sys/mman.h
- Adds a ifdef around the memory manager mutex (using _MMAN_MUTEX_)

### DIFF
--- a/examples/library/test_manycore_alignment/main.c
+++ b/examples/library/test_manycore_alignment/main.c
@@ -33,7 +33,6 @@
 #include <bsg_manycore_npa.h>
 #include <bsg_manycore_config_pod.h>
 #include <stdio.h>
-#include <sys/ioctl.h>
 #include <unistd.h>
 #include <string.h>
 #include <time.h>
--- a/examples/library/test_manycore_eva/main.c
+++ b/examples/library/test_manycore_eva/main.c
@@ -33,7 +33,6 @@
 #include "test_manycore_parameters.h"
 
 #include <stdio.h>
-#include <sys/ioctl.h>
 #include <unistd.h>
 #include <string.h>
 #include <time.h>
@@ -166,7 +165,7 @@ int test_manycore_eva () {
                         tgt_epa = rand() % DRAM_EPA_SIZE; // Small, but we'll deal.
 
                         tgt_x = (rand() % (dim_x - 1)) + origin_x;
-                        tgt_y = hb_mc_config_get_dram_y(config);
+                        // tgt_y = hb_mc_config_get_dram_y(config);
 
                         eva =   (  (tgt_epa & DRAM_STRIPE_MASK)
                                  | (tgt_x << dram_x_offset)
--- a/examples/library/test_struct_size/main.c
+++ b/examples/library/test_struct_size/main.c
@@ -26,7 +26,6 @@
 // SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #include <stdio.h>
-#include <sys/ioctl.h>
 #include <unistd.h>
 #include <string.h>
 #include <time.h>
--- a/libraries/bsg_manycore_memory_manager.h
+++ b/libraries/bsg_manycore_memory_manager.h
@@ -28,7 +28,9 @@
 
 namespace awsbwhal {
         class MemoryManager {
-                std::mutex mMemManagerMutex;
+                #ifdef _MMAN_MUTEX_
+                  std::mutex mMemManagerMutex;
+                #endif
                 std::list<std::pair<uint64_t, uint64_t> > mFreeBufferList;
                 std::list<std::pair<uint64_t, uint64_t> > mBusyBufferList;
                 const uint64_t mSize;
--- a/examples/library/test_manycore_credits/main.c
+++ b/examples/library/test_manycore_credits/main.c
@@ -33,7 +33,6 @@
 #include <inttypes.h>
 #include <cl_manycore_regression.h>
 #include <stdio.h>
-#include <sys/ioctl.h>
 #include <unistd.h>
 #include <string.h>
 #include <time.h>
--- a/examples/library/test_manycore_dmem_read_write/main.cpp
+++ b/examples/library/test_manycore_dmem_read_write/main.cpp
@@ -32,7 +32,6 @@
 #include <cinttypes>
 #include <type_traits>
 #include <stdio.h>
-#include <sys/ioctl.h>
 #include <unistd.h>
 #include <string.h>
 #include <time.h>
--- a/examples/library/test_manycore_dram_read_write/main.c
+++ b/examples/library/test_manycore_dram_read_write/main.c
@@ -31,7 +31,6 @@
 #include <bsg_manycore_coordinate.h>
 #include <bsg_manycore_vcache.h>
 #include <stdio.h>
-#include <sys/ioctl.h>
 #include <unistd.h>
 #include <string.h>
 #include <time.h>
@@ -121,7 +120,7 @@ int test_manycore_dram_read_write() {
                                 if (err != HB_MC_SUCCESS) {
                                         bsg_pr_err("%s: failed to read A[%d] "
                                                    "from DRAM coord(%d,%d) @ 0x%08" PRIx32 "\n",
-                                                   i,
+                                                   __func__, i,
                                                    hb_mc_coordinate_get_x(dram_coord),
                                                    hb_mc_coordinate_get_y(dram_coord),
                                                    BASE_ADDR + i);
--- a/examples/library/test_manycore_packets/main.c
+++ b/examples/library/test_manycore_packets/main.c
@@ -31,7 +31,6 @@
 #include <bsg_manycore_tile.h>
 #include <bsg_manycore_printing.h>
 #include <stdio.h>
-#include <sys/ioctl.h>
 #include <unistd.h>
 #include <string.h>
 #include <time.h>
--- a/examples/library/test_manycore_vcache_sequence/main.c
+++ b/examples/library/test_manycore_vcache_sequence/main.c
@@ -33,7 +33,6 @@
 #include <bsg_manycore_npa.h>
 #include <bsg_manycore_config_pod.h>
 #include <stdio.h>
-#include <sys/ioctl.h>
 #include <unistd.h>
 #include <string.h>
 #include <time.h>
@@ -118,7 +117,7 @@ int test_manycore_vcache_sequence() {
                                         bsg_pr_test_info("\n%s: mismatch @ index %d: "
                                                          "wrote 0x%08" PRIx32 " -- "
                                                          "read 0x%08" PRIx32 ": @ 0x%08" PRIx32 "\n",
-                                                         __func__, i, i, write_data, read_data,
+                                                         __func__, i, write_data, read_data,
                                                          hb_mc_npa_get_epa(&npa));
                                 }
                         
--- a/examples/library/test_printing/main.c
+++ b/examples/library/test_printing/main.c
@@ -44,7 +44,6 @@
 /********************************************************************/
 
 #include <stdio.h>
-#include <sys/ioctl.h>
 #include <unistd.h>
 #include <string.h>
 #include <time.h>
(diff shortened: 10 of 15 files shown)


dpetrisko:
I would separate out the header changes that are RISC-V agnostic and open a different PR for the hammerparrot platform

examples/library/test_struct_size/main.c
```diff
@@ -36,6 +35,7 @@
 #include <bsg_manycore_printing.h>
 #include <cl_manycore_regression.h>
 
+// This test seems to be superseded by test_packet. Do we need it?
```
dpetrisko, line 38: Not sure how @drichmond feels about it, but my preference for PRs is to add things like this as GH comments rather than code comments, and then reviewers can determine whether to open another issue/PR. Otherwise they tend to get lost in the noise

libraries/platforms/hammerparrot-vcs/software/include/argp.h
```diff
@@ -0,0 +1,278 @@
+/*
```
dpetrisko, line 1: LICENSE?

dpetrisko: for all copied files

dpetrisko: or point to them in the toolchain rather than copy

sripathi-muralitharan: Added the LICENSE information. For endian.h and features.h, I just used pointers to the original files. For the argp header, I put a pointer in the file as a comment. It comes from the linux newlib so there is no equivalent in the actual newlib+dramfs that we use.

sripathi-muralitharan:
To run upto compilation of the library files, I think we require `hardware.mk`, `compilation.mk` and `link.mk`. So, I thought I might as well add the other two. `hardware.mk`, `link.mk`, `execution.mk` and `library.mk` are exact copies from the files in bigblade-vcs. Only `compilation.mk` is changed.

examples/library/test_manycore_eva/main.c
```diff
@@ -166,7 +165,7 @@ int test_manycore_eva () {
                         tgt_epa = rand() % DRAM_EPA_SIZE; // Small, but we'll deal.
 
                         tgt_x = (rand() % (dim_x - 1)) + origin_x;
-                        tgt_y = hb_mc_config_get_dram_y(config);
+                        // tgt_y = hb_mc_config_get_dram_y(config);
```
sripathi-muralitharan, line 168: No y coordinates required for DRAM pods?

examples/library/test_struct_size/main.c
```diff
@@ -26,7 +26,6 @@
 // SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
```
sripathi-muralitharan, line 27: Is this test superseded by test_packet?

examples/library/test_struct_size/main.c
sripathi-muralitharan: That makes sense. So I reverted to using GH comments.

dpetrisko:
Sweet, my feedback is addressed, I’ll leave the rest to @drichmond and @mrutt92

dpetrisko:
Can we remove the hammerparrot platform and get the header removals / additions in?

drichmond:
Is that to us or @sripathi-muralitharan?

dpetrisko:
@sripathi-muralitharan for the separation, @drichmond for the review :P

sripathi-muralitharan:
Removed the platform makefiles and hardware. Only the header file additions and removals now remain.

drichmond:
What about the argp header, and changes to .gitignore?

sripathi-muralitharan:
`argp.h` is required for the software port. Its used at some point. I have pointed to the file from where I take it and its an exact copy from that location. Unfortunately, this header does not exist in the newlib+dramfs toolchain, so I had to copy it over from the linux version.

I can get rid of the .gitignore

drichmond:
But that's part of the platform

sripathi-muralitharan:
Wait, you don't want any of the headers at all?

examples/library/test_manycore_alignment/main.c
```diff
@@ -164,7 +163,7 @@ int test_manycore_alignment() {
                                                  dram_coord_y,
                                                  BASE_ADDR + 1);
                         write_data = rand();
-                        err = hb_mc_manycore_write32(mc, &npa, &write_data);
+                        err = hb_mc_manycore_write32(mc, &npa, write_data);
```
drichmond, line 166: Is this just an outdated test? What went wrong here?

drichmond:
`git rm libraries/platforms/hammerparrot-vcs/`

Nothing in that directory should be in this PR

examples/library/test_manycore_alignment/main.c
sripathi-muralitharan: The `hb_mc_manycore_write*` functions call the `hb_mc_manycore_write` functions by passing the address of the third argument (i,e write_data here). Passing the third argument as an address, would make the call to write a pointer to a pointer. Is that the intention? If it is, I am wrong and I will revert it.

drichmond: No, you're probably right. Just confused why this wasn't caught before. Does the test pass?

dpetrisko:
>argp.h is required for the software port. Its used at some point. I have pointed to the file from where I take it and its an exact copy from that location. Unfortunately, this header does not exist in the newlib+dramfs toolchain, so I had to copy it over from the linux version.

I guess the question is if we want to use different headers for different platforms. Totally up to you, but it makes sense to me to include this header in the main library so that everything is using the same C++ codebase

examples/library/test_manycore_alignment/main.c
sripathi-muralitharan: Yes.

drichmond:
> > argp.h is required for the software port. Its used at some point. I have pointed to the file from where I take it and its an exact copy from that location. Unfortunately, this header does not exist in the newlib+dramfs toolchain, so I had to copy it over from the linux version.
> 
> I guess the question is if we want to use different headers for different platforms. Totally up to you, but it makes sense to me to include this header in the main library so that everything is using the same C++ codebase

If that's the case, then the main libraries directory, or the examples directory are better places. And it should be a separate PR.

examples/library/test_manycore_alignment/main.c
drichmond: huh...

examples/library/test_manycore_eva/main.c
drichmond: There is, but it's handled in a different way. This is @mrutt92's least favorite test

examples/library/test_manycore_alignment/main.c
sripathi-muralitharan: It passes with the changes I have. I can check it without the changes xP

dpetrisko:
> > > argp.h is required for the software port. Its used at some point. I have pointed to the file from where I take it and its an exact copy from that location. Unfortunately, this header does not exist in the newlib+dramfs toolchain, so I had to copy it over from the linux version.
> > 
> > 
> > I guess the question is if we want to use different headers for different platforms. Totally up to you, but it makes sense to me to include this header in the main library so that everything is using the same C++ codebase
> 
> If that's the case, then the main libraries directory, or the examples directory are better places. And it should be a separate PR.

Yup agree with both

examples/library/test_manycore_alignment/main.c
sripathi-muralitharan: Both versions of the test (with the changes I proposed and without them) work. So, I am not sure what is the course of action of here. Keep it the same?

dpetrisko: Oh an always passing test, that’s a classic. 

Course of action is revert the change and file an issue

sripathi-muralitharan: Opened #713 for this. 

Also found that the test_pod_iteration library test fails due to an `Invalid Input` error raised by the `hb_mc_platform_init` function. Opened #712 for that

examples/library/test_struct_size/main.c
mrutt92: lol - whoops. forgot this exists.

mrutt92: in my defense, what a non-descript name?

mrutt92: I think we can nuke this.

examples/library/test_manycore_eva/main.c
mrutt92: This test suuuuuuuuuuuxxxxxxxxxx. When it fails, 99 times out of 100 the solution is to fix the test instead of any RTL or library code.

mrutt92: We use these now with pods https://github.com/bespoke-silicon-group/bsg_replicant/blob/c06120ac2dd477aedb88973a709bb1d81ccc8101/libraries/bsg_manycore_config_pod.h#L269.

mrutt92 (approved):
PR looks good to me. I might change the name of this PR to "Removes cruft headers and macro guards mutex" and edit the description with bullets saying what it does. The current name suggests something... bigger.

examples/library/test_manycore_eva/main.c
sripathi-muralitharan: Ahh okay.

examples/library/test_struct_size/main.c
sripathi-muralitharan: Should I get rid of it then before we merge?

dpetrisko: separate PR I think

taylor-bsg:
I suggest having a link from the EVA code to this test as something that
needs to be updated if EVA map is changed.

dpetrisko:
This ready to go? Possibly with new issue opening for this:

>I suggest having a link from the EVA code to this test as something that
needs to be updated if EVA map is changed.

sripathi-muralitharan:
New issue or just a link to https://github.com/bespoke-silicon-group/bsg_manycore/blob/ci_bigblade/v/bsg_manycore_eva_to_npa.v? 
There already seems to be a link back to the cuda-lite library code in the SV file.

dpetrisko:
I think michael's suggestion is to add a link to the brittle test as well as the library

libraries/bsg_manycore_memory_manager.h
```diff
@@ -28,7 +28,9 @@
 
 namespace awsbwhal {
         class MemoryManager {
-                std::mutex mMemManagerMutex;
+                #ifdef _MMAN_MUTEX_
+                  std::mutex mMemManagerMutex;
+                #endif
```
drichmond, line 33: is this a standard macro? Or is this something we need to define now?

sripathi-muralitharan: I don't think it is a standard macro. We will need to define it separately. Is library.mk a good spot for this?

dpetrisko: Is there a more basic switch? Like if newlib?

sripathi-muralitharan: This might conflict with something in newlib. I guess, we can be safer and say if BLACKPARROT or something like that. What do you think?

mrutt92: I like the macro how it is.

sripathi-muralitharan:
Are there any other changes required for this PR?

libraries/bsg_manycore_memory_manager.h
mrutt92: Not standard (and doesn't conflict with anything): good
Descriptive (does exactly what the name would imply): good
Does the job: excellent
What's the problem?

sripathi-muralitharan: Well, the original question was if this was a standard macro which you just confirmed it isn't. So, there is no problem.

mrutt92:
I think this fine, let's hammer this parrot!

PR #715, opened 2021-04-07
Update Verilator link.mk

dpetrisko:
- Here I remove -j and rely on the user to do make -j
- We could also do -j$(NUM_THREADS) or something
- -j by itself starts infinite jobs and crashes my laptop


mrutt92:
strong reject

mrutt92:
jk

mrutt92 (approved):

drichmond:
I used to have an NPROC (or similar) variable for VCS compilation. It would determine the number of threads in the system and then divide by four. 

Should be in the git history somewhere. 

But if you don't want to track it down, LGTM

taylor-bsg (approved):

dpetrisko:
good to go?

PR #717, opened 2021-04-20
Adding _DEFAULT_SOURCE macro

dpetrisko:
Newer versions of glibc require this macro to be set alongside _BSD_SOURCE and _GNU_SOURCE: https://stackoverflow.com/questions/29201515/what-does-d-default-source-do


drichmond (approved):

mrutt92:
uh... err... were we suppose to merge this?

dpetrisko:
Let me update it then #shipit

dpetrisko:
Good to go

mrutt92:
awesome possum. and you promise that this will 100% not breaking everything, right? ;-)

dpetrisko:
It's only 84 files, how much damage could I do? :D

mrutt92:
Good enough for me!

PR #719, opened 2021-04-27
Bigblade regression header refactor

mrutt92:
* Moves functionality in cl_manycore_regression.h into a new regression shared library
* Moves command line parsing into platform specific source

### DIFF
--- a/libraries/platforms/bigblade-vcs/bsg_manycore_regression_platform.c
+++ b/libraries/platforms/bigblade-vcs/bsg_manycore_regression_platform.c
@@ -1,6 +1,6 @@
 #include <stdint.h>
 #include <unistd.h>
-#include "bsg_manycore_regression.h"
+#include <bsg_manycore_regression.h>
 // Given a string, determine the number of space-separated arguments
 static int get_argc(char * args){
         char *cur = args, prev=' ';
--- a/examples/cuda/test_dram_load_store/Makefile
+++ b/examples/cuda/test_dram_load_store/Makefile
@@ -68,10 +68,6 @@ CXXFLAGS += -std=c++11 $(FLAGS)
 # compilation.mk defines rules for compilation of C/C++
 include $(EXAMPLES_PATH)/compilation.mk
 
-# Specify any header file dependencies
-main.o: INCLUDES += -I$(EXAMPLES_PATH)
-main.o: $(EXAMPLES_PATH)/cl_manycore_regression.h
-
 ###############################################################################
 # Host code link flags and flow
 ###############################################################################
--- a/examples/cuda/test_dram_load_store/main.c
+++ b/examples/cuda/test_dram_load_store/main.c
@@ -45,7 +45,7 @@
 #include <unistd.h>
 #include <sys/ioctl.h>
 #include <stdio.h>
-#include <cl_manycore_regression.h>
+#include <bsg_manycore_regression.h>
 
 #define ALLOC_NAME "default_allocator"
 
@@ -162,15 +162,4 @@ int kernel_dram_load_store(int argc, char **argv) {
         return HB_MC_SUCCESS;
 }
 
-#ifdef VCS
-int vcs_main(int argc, char ** argv) {
-#else
-int main(int argc, char ** argv) {
-#endif
-        bsg_pr_test_info("test_dram_load_store Regression Test \n");
-        int rc = kernel_dram_load_store(argc, argv);
-        bsg_pr_test_pass_fail(rc == HB_MC_SUCCESS);
-        return rc;
-}
-
-
+declare_program_main("test_dram_load_store", kernel_dram_load_store);
--- a/examples/cuda/test_empty_parallel/Makefile
+++ b/examples/cuda/test_empty_parallel/Makefile
@@ -68,10 +68,6 @@ CXXFLAGS += -std=c++11 $(FLAGS)
 # compilation.mk defines rules for compilation of C/C++
 include $(EXAMPLES_PATH)/compilation.mk
 
-# Specify any header file dependencies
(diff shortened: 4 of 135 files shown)


drichmond:
In general, good. I think I agree with the focus of the PR, which is basically: Manage per-platform main (or lack thereof) and argument parsing. 

declare_program_main feels funky to me, but I understand why. Just make sure there's a comment in the source where it's used. 

I think I'm ok with the float compare/generation stuff in a .c file. 

I worry about new tests that will define new arguments so that sweeps can be performed. Does that need to live in a c file, or is a .h file ok?

drichmond:
OK, I think I see how new arguments would work. It's not handled in "main", it's handled in the test function.

mrutt92:
> declare_program_main feels funky to me, but I understand why. Just make sure there's a comment in the source where it's used.

It's taking code that was, in practice, copy-and-pasted for every single test and instead moving it into three or four header files (one for each platform). How's it more clunky than what was there before? Right now I've only applied to test_rom, but there is similar code to what was removed from test_rom in every test.

> I worry about new tests that will define new arguments so that sweeps can be performed. Does that need to live in a c file, or is a .h file ok?

Those tests can do whatever they want. Argc and argv are available to each test and I can think of three reasonable options:

1. They can just assume positional arguments / write their own command line parser like 99% of other C/C++ programs do. I've done this myself a couple of times because I find argp clunky.
2. They can look at examples for argp that we've written ourselves and make their own.
3. They can (2) and submit a PR that adds it to the library.

drichmond:
> > declare_program_main feels funky to me, but I understand why. Just make sure there's a comment in the source where it's used.
> 
> It's taking code that was, in practice, copy-and-pasted for every single test and instead moving it into three or four header files (one for each platform). How's it more clunky than what was there before? Right now I've only applied to test_rom, but there is similar code to what was removed from test_rom in every test.

Yep. As I said, feels clunky but I understand why. I don't have a better suggestion. In a header is good. 

> 
> > I worry about new tests that will define new arguments so that sweeps can be performed. Does that need to live in a c file, or is a .h file ok?
> 
> Those tests can do whatever they want. Argc and argv are available to each test and I can think of three reasonable options:
> 
> 1. They can just assume positional arguments / write their own command line parser like 99% of other C/C++ programs do. I've done this myself a couple of times because I find argp clunky.
> 2. They can look at examples for argp that we've written ourselves and make their own.
> 3. They can (2) and submit a PR that adds it to the library.

Yep.

mrutt92:
>>> declare_program_main feels funky to me, but I understand why. Just make sure there's a comment in the source where it's used.

>>It's taking code that was, in practice, copy-and-pasted for every single test and instead moving it into three or four header files (one for each platform). How's it more clunky than what was there before? Right now I've only applied to test_rom, but there is similar code to what was removed from test_rom in every test.

>Yep. As I said, feels clunky but I understand why. I don't have a better suggestion. In a header is good.

For the sake of argument, the other option is force each test to define a function of a certain name, maybe call it "test_main," and then we just have cosm_main and main (depending on the platform) call it. If you want to customize the test name that's printed at the start of the regression they can also define a special string, call it "test_name," that we print out. Personally, I think the macro sounds like less work for the test writer.

dpetrisko:
ya love to see it: 
image

mrutt92:
@drichmond this is ready for review.

mrutt92:
@sripathi-muralitharan let me know if this works with what you're doing on black parrot.

sripathi-muralitharan:
@mrutt92 I think this is great! This separation seems to clearly define what is used by the CUDA-lite program itself and things like defining what the platform needs like pulling arguments from VCS, which is exactly what I wanted for BlackParrot as well.

mrutt92:
@drichmond let's approve this and get it merged in.

libraries/platforms/bigblade-vcs/bsg_manycore_regression_platform.c
```diff
@@ -0,0 +1,66 @@
+#include <stdint.h>
+#include <unistd.h>
+#include "bsg_manycore_regression.h"
```
drichmond, line 3: Angle brackets

drichmond:
Looks good to me. One question I had: bsg_manycore_regression.c -- should it be compiled as part of the shared library or as part of the high-level application?

If it's in the shared library every platform needs to support argparse, but if it is part of the application only the test needs to support/include argparse.

mrutt92:
Hmm... good question. Frankly, I would get rid of argparse completely. The command line arguments passed to c_args are typically so simple that I'm not sure what argparse actually does for us.

drichmond:
Nice things

drichmond:
I agree with your thoughts about argparse, for these tests. But I think it would be good to keep for the long run. Helps with generality.

Influenced by this, I think I would rather see it linked in the final step of each program. Then people can opt to not use argparse, and we can remove it gradually in the future if we so choose.

mrutt92:
What nice things? Does pythorch use arparse in some fashion that I'm not aware of? AFAICT we use argparse to do something that the following lines of code would accomplish:

```c
char *exe, *rv32;
exe = argv[0];
if (argc > 1) {
  rv32 = argv[1];
}
```

mrutt92:
> I agree with your thoughts about argparse, for these tests. But I think it would be good to keep for the long run. Helps with generality.

I think it does the opposite. Almost nothing is more general than argv. Let's be clear here, if someone wanted to write a test where they want to have a different cli, and they wanted to argparse, they would benefit very little from our existing code because they would need to (1) define their own argp struct, (2) define their own options struct, and (3) define their own parsing function. Basically, they need to start with argp from scratch. 

That's my opinion, but maybe you can expand on what generality argp buys us?

drichmond (approved):

PR #720, opened 2021-04-30
WIP: Bigblade unit tests

mrutt92:
* This introduces a first unit test, which I hope to be a template for others, that reads from all ways of a target L2 set.


mrutt92:
@drichmond what do you think? Unlike in most situations, input is actually welcome in this case :-)

PR #724, opened 2021-05-03
HammerBlade C++ wrapper

mrutt92:
* Simple wrapper for CUDA-Lite host code.


mrutt92:
@drichmond you might not think this is so exciting once you look at it, but here it is.

drichmond:
I think this is awesome. With some extra documentation in the API (to match our C libraries) I think we should go ahead and merge this.

I would also rename the associated test to signal that it is testing the associated C++ api.

PR #725, opened 2021-05-04
Created new platform for tapeout design

drichmond:
Work in progress (waiting for chip RTL) but functionally complete minus some error checks.

Changed VCS compilation flow for the chip tapeout. Previously we compiled the RTL for each program, but compiling the chip RTL slow and apparently error prone.

This PR adds a new platform, tapeout-vcs, which compiles the simulation executable separately from the program. Now each machine has several different flavors of executable depending on which feature is required (SAIF, profilers, nothing, or waveforms). The program is now compiled as a .so file, which is dynamically loaded (dlopen, dlsym, dlerror, dclose) during simulation.

The program should now only be run through the make interface (this was typical before). The same rules work as in bigblade.


dpetrisko:
Is this the plan for future platforms or just a one-off for chip RTL?

drichmond:
TBD.

mrutt92:
TBD = prepare yourself.

drichmond:
If we make this standard, it will only be standard for VCS Platforms. Vastly reduces compilation cycles

dpetrisko:
Does Verilator not compile the binary in the same way?

drichmond:
No. This actually makes VCS compilation more like verilator from the 10000 foot level.

VCS only analyzes the verilog in the first step (vlogan). When we actually run VCS to do the linking step we "compile" the verilog. This apparently can take a long time for netlists and is complex (?) so we're simplifying the build process to a single machine binary.

mrutt92 (approved):

PR #730, opened 2021-05-18
Bigblade -O3

mrutt92:


drichmond (approved):
Sure, but why?

drichmond:
LGTM

mrutt92:
1. It should have always been this way. IMO, the fact that it's been -O0 is a bug.
2. When running host software on RISCV simulation platforms like Dromajo and BlackParrot, every instruction matters.

drichmond:
Yep, saw the slack thread. -Os might be worth trying, too.

mrutt92:
To be clear, I meant every instruction executed. Sripathi tried -Os and saw that the code size was not much improved over -O3.

PR #738, opened 2021-07-02
Tinkering with RISC-V compilation flow

drichmond:


mrutt92:
add a comment regarding the issue we tripped on and then I think this is good to go

mrutt92 (approved):

taylor-bsg:
The comments kind of justify at a technical level why there is additional complexity, but the makefile doesn't really address what functionality is being supplied and how the user is supposed to interact with this functionality?

drichmond:
Sure, I'll add it. 

But for the sake of posterity on github:

Background: riscv.mk is a way to compile RISC-V code in bsg_replicant, so that the application code can co-exist with the binary, and remove the need to file two PRs when adding applications. riscv.mk was originally in the "baseline" repository, but maintaining that repository is difficult when there's a lot of code churn, so we brought riscv.mk here and deprecated baseline. In short, it eases the development of kernels, and explaining how to develop kernels.

This PR: 

When compiling RISC-V code we've often said "I would like to compile part of my program with LLVM and part of my program with GCC" because sometimes LLVM does better than GCC. This behavior is not currently an option in bsg_manycore, because the makefiles require CLANG=1 to be set, which compiles ALL application objects with LLVM, or none of them. 

In a typical flow you can change your C/C++ compiler by setting CC or CXX like this:

Globally:
`CC=clang`
`CXX=clang++`

Per-object:
`foo.o: CC=clang`
`foo.o: CXX=clang++`

The latter is nice because you can specify how each object gets compiled (or use a default). This example works in x86. This PR provides this functionality for RISC-V:

Globally:
`RISCV_CC=$(RISCV_GCC)`
`RISCV_CXX=$(RISCV_CLANGXX)`

Per-object:
`foo.rvo: RISCV_CC=$(RISCV_GCC)`
`foo.rvo: RISCV_CXX=$(RISCV_CLANGXX)`

And this will compile RISC-V binaries using clang or gcc using the specified compiler. We use RISCV_CC and RISCV_CXX because CC and CXX are for host (usually x86) compilation.

The complexity arises because of clang in bsg_manycore. Clang should be able to compile HB RISC-V code as a drop-in replacement for bsg_manycore's RISC-V GCC and recognize HB attributes like `bsg_remote_attr`. If it could, this PR would not be required, because we could just follow the first practice and compile each object independently by specifying RISCV_CLANG or RISCV_GCC.

However, there's some issue with clang handing off arguments to llvm/llc, and clang doesn't pass `-mcpu=hb-rv32` to llc, which means that loads to a pointer that are labeled with `bsg_remote_attr` don't get weighted correctly. (This is my interpretation of the issue, and is consistent with the comment [here](https://github.com/bespoke-silicon-group/bsg_manycore/blob/3a05e000aef1fba896f756e9836c9e002e5a06f7/software/mk/Makefile.builddefs#L208)). As a result, loads are not hoisted and code is sub-optimal using LLVM.

This is why bsg_manycore requires CLANG=1. It changes the compilation into multiple steps (clang -> LLVM IR -> llc -> optimized assembly -> clang -> object file) using targets that only get enabled/disabled with CLANG=1 (see [here](https://github.com/bespoke-silicon-group/bsg_manycore/blob/3a05e000aef1fba896f756e9836c9e002e5a06f7/software/mk/Makefile.builddefs#L200)).

This PR wraps RISCV_GCC and RISCV_CLANG as makefile functions, so that RISCV_CLANG looks like a single command instead of multiple steps.

(I ran into this while working with Xingyao on credits during tapeout and I pulled it forward to here during close-out)

examples/cuda/riscv.mk
```diff
@@ -49,41 +61,140 @@ RISCV_GNU_PATH   := $(RISCV_TOOLS_PATH)/riscv-install
...
+$(_RISCV_CLANG) $(RISCV_CFLAGS) $(RISCV_DEFINES) $(RISCV_INCLUDES) -c $< -o $@.ll -S -emit-llvm |& tee $*.rvo.ll.log && \
+$(RISCV_LLVM_LLC) $(RISCV_LLVM_LLC_FLAGS) $@.ll -o $@.S |& tee $*.rvo.S.log && \
+$(_RISCV_CLANG) $(RISCV_CFLAGS) $(RISCV_DEFINES) $(RISCV_INCLUDES) -c $@.S -o $@ |& tee $*.rvo.log
+endef
+
+# Set the default RISC-V Compilers.
 RISCV_CXX ?= $(RISCV_GXX)
 RISCV_CC  ?= $(RISCV_GCC)
 
+# To chose between GCC and CLANG set RISCV_CC to RISCV_GCC or RISCV_CLANG, respectively.
```
mrutt92, line 178: I don't think the use of ", respectively" is needed here (or correct?)

examples/cuda/riscv.mk
```diff
@@ -49,41 +61,140 @@ RISCV_GNU_PATH   := $(RISCV_TOOLS_PATH)/riscv-install
...
+# multiple steps:
+#   1. Clang (.c/.cpp -> .ll (LLVM IR))
+#   2. llc (.ll (LLVM IR) -> .s)
+#   3. gcc/clang (.s -> .rvo)
+
+# This is problematic if we want to selectively compile objects with
+# clang/LLVM because it adds intermediate targets that GCC does not
+# have. AFAIK there is no easy way to express two different
+# compilation paths in a single run of make. If clang correctly passed
+# the arguments to lcc and HB RV32 objects could be optimally comiled
```
mrutt92, line 134: comiled -> compiled

sampsyo:
This is a heroic effort! To return to "single Clang invocation" convenience, one place I might start looking is the Clang driver's handling of `-mcpu` and its translation to internal feature flags:
https://github.com/llvm/llvm-project/blob/f1cbea3e527547fc08c55235c11970a8d9f2637e/clang/lib/Driver/ToolChains/Arch/RISCV.cpp#L469-L480

examples/cuda/riscv.mk
drichmond: Fixed

examples/cuda/riscv.mk
drichmond: I think it is

mrutt92 (approved):

PR #739, opened 2021-07-07
Bigblade

drichmond:
* Add support for multi-pod machines
* Switch to new application-as-shared-object flow for simulation
* Move tests into independent directories
* Add support for Credit limit CSR
* Add bigblade-vcs platform as default platform
* Add tapeout-vcs platform for running on a simv in an external repository
* Add RISC-V compilation flow with support for per-object gcc / LLVM compilation


mrutt92 (approved):

drichmond:
285 changes are from moving tests into independent directories
15 are from unit tests
The remaining changes are mostly to support pods, DMA, and the two new platforms.

PR #743, opened 2021-07-28
BFS+GUPs benchmarks

mrutt92:


scripts/py/dramsim3_utils.py
```diff
@@ -0,0 +1,113 @@
...
+    with open(filename_i, "r") as jsonfile:
+        lines = [line for line in jsonfile]
+
+    for (lno, line) in enumerate(lines):
+        # if this is the last line
+        # replace the ',' with ']'
+        if lno == len(lines)-1:
+            line = line.replace(',\n',']\n')
+        # write it out
+        jsonfile_o.write(line)
```
drichmond, line 16: Little confused why this is necessary

I thought we had a fix that closed out the json file. Is this the same issue?

machines/bigblade_pod_X1Y1_ruche_X16Y8_hbm/Makefile.machine.include
```diff
@@ -0,0 +1,135 @@
+# Copyright (c) 2019, University of Washington All rights reserved.
```
drichmond, line 1: I would separate this into another PR

libraries/features/dma/simulation/dramsim3.mk
```diff
@@ -64,6 +64,7 @@ $(LIBRARIES_PATH)/features/dma/simulation/libdramsim3.so: CXXFLAGS += -I$(BASEJU
 $(LIBRARIES_PATH)/features/dma/simulation/libdramsim3.so: CXXFLAGS += -I$(BASEJUMP_STL_DIR)/imports/DRAMSim3/ext/fmt/include
 $(LIBRARIES_PATH)/features/dma/simulation/libdramsim3.so: CXXFLAGS += -DFMT_HEADER_ONLY=1
 $(LIBRARIES_PATH)/features/dma/simulation/libdramsim3.so: CXXFLAGS += -DBASEJUMP_STL_DIR="$(BASEJUMP_STL_DIR)"
+$(LIBRARIES_PATH)/features/dma/simulation/libdramsim3.so: CXXFLAGS += -DBLOOD_GRAPH=1
```
drichmond, line 67: Always on?

examples/cuda/gups/.gitignore
```diff
@@ -0,0 +1 @@
+*_run/
```
drichmond, line 1: Newline plz

examples/cuda/bfs/.gitignore
```diff
@@ -0,0 +1,6 @@
+input-analyze
+input-gen
+input-degree-list
+inputs/
+*tgy/
+stack.info.*
```
drichmond, line 6: Newline plz

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "examples/cuda/gups/hammerblade-helpers"]
+	path = examples/cuda/gups/hammerblade-helpers
+	url = git@github.com:mrutt92/hammerblade-helpers
```
drichmond, line 3: I feel like this should be somewhere else in the directory tree long-term, but since you're the only one using it right now, LGTM

scripts/py/vcache_utils.py
```diff
@@ -0,0 +1,50 @@
+import pandas as pd
+
```
drichmond, line 2: If you ever want it, I have the CudaStatTag class popped out in another branch

scripts/py/dramsim3_utils.py
mrutt92: I was surprised too, looks like it didn't make it in to the .tag.json.

drichmond: @marcusnchow Do you remember anything about this? I thought we fixed it.

marcusnchow: [This](https://github.com/bsg-external/DRAMsim3/blob/1b8828cd293077e321a028310a8b938031b21ff5/src/dram_system.cc#L73) was the fix that i made that replaces the last ']' with a ','

Not sure if this is the same file that this script is working with?
Or do you know if BaseDRAMSystem::PrintStats() is being called?

.gitmodules
mrutt92: Yeah there's a PR that was never merged.

scripts/py/vcache_utils.py
mrutt92: good to know. let this be the start of a stats scripting directory in replicant? honestly, I just use It to reformat the csv into a table I can more easily manage in excel.

libraries/features/dma/simulation/dramsim3.mk
mrutt92: Hrm. Probably not? We could make a version of libdramsim3 with blood graphs compiled in and then only link when profiling. That would more or less match what we do for the vanilla blood graphs right? With the tracer?

machines/bigblade_pod_X1Y1_ruche_X16Y8_hbm/Makefile.machine.include
mrutt92: OK, see #744

.gitmodules
mrutt92: See #724

machine.mk
```diff
@@ -38,7 +38,7 @@ endif
 
 # To switch machines, simply switch the path of BSG_MACHINE_PATH to
 # another directory with a Makefile.machine.include file.
-BSG_MACHINE_PATH ?= $(BSG_F1_DIR)/machines/pod_X1Y1_ruche_X16Y8_hbm
+BSG_MACHINE_PATH ?= $(BSG_F1_DIR)/machines/bigblade_pod_X1Y1_ruche_X16Y8_hbm
```
drichmond, line 41: Is this intentional? I thought if it was, it would go in #744

libraries/features/dma/simulation/dramsim3.mk
drichmond: Do DRAMSim3 bloodgraphs slow down sim?

.gitmodules
drichmond: Wondered. Do you want that merged, or keep this as a submodule for now?

scripts/py/vcache_utils.py
drichmond: Any scripts that are generally useful for manycore should go in manycore repository, but it seems like you're still iterating on this so I'm fine with this for now.

scripts/py/dramsim3_utils.py
mrutt92: OK, my bad. My helper library uses a smart pointer and it calls DPI finish functions for the device after the finally blocks in the SV. I've added calls to explicitly perform HB cleanup code.

machine.mk
mrutt92: removed. not intentional.

scripts/py/vcache_utils.py
mrutt92: well you know what they say about rules of thumb... some are made to be broken.

mrutt92: you could make the argument that this more generally useful for cuda than for manycore.

.gitmodules
mrutt92: I don't know. We should talk about it.

libraries/features/dma/simulation/dramsim3.mk
mrutt92: I'm going to go ahead and say "probably"

scripts/py/vcache_utils.py
drichmond: Wait, thumbs are meant to be broken?

You heard the discussion in meeting this morning. Replicant is a public repo but manycore is the focus repository. If it's generally useful, and it's finalized it should go in manycore.

.gitmodules
drichmond: Urgently?

libraries/features/dma/simulation/dramsim3.mk
drichmond: Then we should come up with a way of turning it on and off, unless we know otherwise

.gitmodules
mrutt92: nah, this is something we can fix later.

mrutt92: I'm more concerned with getting this merged so it doesn't go stale.

scripts/py/vcache_utils.py
mrutt92: Yeah I heard that... but I think when it comes to data post-processing from an experiment I think most researchers are particular enough that they'll have their own flow anyway. That's why it's great that are stats infrastructure within the simulation just dumps out a CSV.

drichmond: SGTM. Then let's move the scripts directory inside bfs or gups since it's your flow at the moment

scripts/py/dramsim3_utils.py
drichmond: Cool

scripts/py/vcache_utils.py
mrutt92: No. This is shared code between the two benchmarks, I'll likely reuse it for more, and I think it's OK for replicant to have a common scripts directory.

libraries/features/dma/simulation/dramsim3.mk
mrutt92: Maybe I can tackle that in a separate PR. It's not super integral to this one.

scripts/py/vcache_utils.py
mrutt92: But if I can always make another submodule.

mrutt92: well instead I just moved it into the one that already exists.

mrutt92:
@drichmond your feedback has been addressed, merge?

drichmond (approved):

drichmond:
Yeah, LGTM

PR #750, opened 2021-09-23
HW Barrier API

mrutt92:
Relies on https://github.com/bespoke-silicon-group/bsg_manycore/pull/583


examples/cuda/hw_barrier/Makefile
```diff
@@ -0,0 +1,141 @@
...
+# BSG_MANYCORE_KERNELS is a list of manycore executables that should
+# be built before executing.
+
+BSG_MANYCORE_KERNELS ?= kernel.riscv
+RISCV_TARGET_OBJECTS += kernel.rvo
+RISCV_INCLUDES  +=
+RISCV_CCPPFLAGS += -D__KERNEL__
+
+TILE_GROUP_DIM_X ?= 16
+TILE_GROUP_DIM_Y ?= 8
```
drichmond, line 94: Are these still necessary with the new HW barrier?

(I'm not asking why they're not gone - just if there's some (cruft?) reason we still need them)

hardware/hardware.mk
```diff
@@ -157,7 +157,7 @@ $(BSG_MACHINE_PATH)/bsg_bladerunner_configuration.rom: $(BSG_MACHINE_PATH)/Makef
 	@echo $(call dec2bin,$(BSG_MACHINE_HOST_COORD_Y))      >> $@.temp
 	@echo $(call dec2bin,$(BSG_MACHINE_NOC_COORD_X_WIDTH)) >> $@.temp
 	@echo $(call dec2bin,$(BSG_MACHINE_NOC_COORD_Y_WIDTH)) >> $@.temp
-	@echo $(call dec2bin,0)                                >> $@.temp
+	@echo $(call dec2bin,$(BSG_MACHINE_RUCHE_FACTOR_X))    >> $@.temp
```
drichmond, line 160: Did you want to print barrier ruche factor or network ruche factor?

drichmond: I realize that all the code reads the machine ruche factor, but doesn't the barrier use the barrier ruche factor?

examples/.gitignore
```diff
@@ -24,3 +24,4 @@ ucli.key
 *.saifgen
 *.json
 dramsim3.txt
+*.a
```
drichmond, line 27: Newline

examples/cuda/hw_barrier/Makefile
taylor-bsg: Also, isn't our intent to specialized against these constants?

drichmond: Yes, I would very much like to specialize. 

But it's nice to have general code.

mrutt92: Is this feedback or a wish?

hardware/hardware.mk
mrutt92: Well flip that question around and ask why there are two different parameters? Can the barrier ruche factor be different from the NOC ruche factor?

drichmond: @tommydcjung

examples/cuda/hw_barrier/Makefile
mrutt92: Yeah... as I think I mentioned in a manycore meeting sometime back, the amoadd barrier could just take the tile group size as arguments through registers. Nobody seemed to be on-board with that, so we went with the current approach.

mrutt92: I think it's OK, how it is. Dynamically sizing a tile group would be cool. but faster code is cooler.

PR #752, opened 2021-10-04
Bigblade verilator

dpetrisko:
Initial cut at Bigblade Verilator support. Currently compiles but does not work. Seems to be a problem with DPI ROM, possibly always returning 0. Putting PR up for tracking and debug


libraries/platforms/bigblade-verilator/hardware/dpi_top.sv
```diff
@@ -0,0 +1,359 @@
+module replicant_tb_top
+  import bsg_manycore_pkg::*;
```
drichmond, line 2: Is this different than the top level in bigblade-vcs?

drichmond:
First pass seems pretty reasonable. I know @mrutt92 hates it but pointing at files in related platforms instead of copying is fine with me, if possib.e

mrutt92:
I do hate that, but it's a precedent set already in the code.

drichmond:
It's one of those things that is on my increasingly long, low-priority, things to change.

mrutt92:
well... it's actually rising on the list as we add more and more platforms.

drichmond:
It's a dynamic priority queue.

mrutt92:
what's kinda interesting about platforms on hb is they've trended towards being "write-once" and unmaintained.

drichmond:
Maintain for bounded time.

To be fair, the current iteration is pretty similar to the dpi-vcs and dpi-verilator platforms. Tommy saved our asses and wrote a common DUT so that we didn't have to port every change and made life simpler.

mrutt92:
Yeah, I guess I'm just brainstorming-out-forum about platforms and if we can break it down into more layers of abstraction so that the code will be easier to maintain. It's tough because it's infrastructure that spans three programming languages (at least).

libraries/platforms/dpi-verilator/library.mk
```diff
@@ -53,8 +53,8 @@ $(PLATFORM_OBJECTS): INCLUDES += -I$(VERILATOR_ROOT)/include/
 $(PLATFORM_OBJECTS): INCLUDES += -I$(BSG_MANYCORE_DIR)/testbenches/dpi/
 $(PLATFORM_OBJECTS): INCLUDES += -I$(BASEJUMP_STL_DIR)/bsg_test/
 
-$(PLATFORM_OBJECTS): CFLAGS    = -std=c11 -fPIC -DVERILATOR $(INCLUDES) -D_GNU_SOURCE -D_BSD_SOURCE -D_DEFAULT_SOURCE
-$(PLATFORM_OBJECTS): CXXFLAGS  = -std=c++11 -fPIC -DVERILATOR $(INCLUDES) -D_GNU_SOURCE -D_BSD_SOURCE -D_DEFAULT_SOURCE
+$(PLATFORM_OBJECTS): CFLAGS    = -std=c11 -fPIC -D_GNU_SOURCE -D_DEFAULT_SOURCE -DVERILATOR $(INCLUDES) -D_BSD_SOURCE -D_DEFAULT_SOURCE
+$(PLATFORM_OBJECTS): CXXFLAGS  = -std=c++11 -fPIC -D_GNU_SOURCE -D_DEFAULT_SOURCE -DVERILATOR $(INCLUDES) -D_BSD_SOURCE -D_DEFAULT_SOURCE
```
drichmond, line 57: Double define of -D_DEFAULT_SOURCE

libraries/platforms/bigblade-verilator/compilation.mk
```diff
@@ -0,0 +1,73 @@
+# Copyright (c) 2019, University of Washington All rights reserved.
+#
```
drichmond, line 2: Is this the same file as in bigblade-vcs?

libraries/platforms/bigblade-verilator/hardware/dpi_top.sv
dpetrisko: removed

libraries/platforms/bigblade-verilator/bsg_manycore_regression_platform.h
```diff
@@ -0,0 +1,41 @@
...
+//        /* your code here */
+//    }
+//    declare_program_main("The name of your test", MyMain)
+//
+#ifdef VCS
+#define declare_program_main(test_name, name)                   \
+    int vcs_main(int argc, char *argv[]) {                      \
+        bsg_pr_test_info("Regression Test: %s\n", test_name);   \
+        int rc = name(argc, argv);                              \
+        bsg_pr_test_pass_fail(rc == HB_MC_SUCCESS);             \
```
drichmond, line 22: I feel like this might be something we should make platform independent

libraries/platforms/bigblade-verilator/compilation.mk
dpetrisko: It is different because we don't generate main.so, which I understand isn't necessary for verilator

libraries/platforms/bigblade-verilator/bsg_manycore_regression_platform.h
dpetrisko: what do you mean by platform independent?

dpetrisko: My understanding is that this is just hacking around making VCS having a main and Verilator needing a main supplied

mrutt92: That understanding is correct. We could define this macro in `bsg_manycore_regression.h` if it's not defined by the platform.

mrutt92: OTOH what would be the default?

mrutt92: also, why is it defined for Verilator and vcs? isn't this platform just for Verilator?

dpetrisko: Oh yeah that's true, could get rid of the ifdef and have the .h be different in each platform

libraries/platforms/bigblade-verilator/compilation.mk
drichmond: Yeah, fair

drichmond:
This is getting close to working. This is the last known issue that remains:

```[78000] %Error: bsg_manycore_link_sif_tieoff.v:45: Assertion failed in TOP.replicant_tb_top.testbench.io_e_tieoff: [BSG_ERROR] Errant fwd packet detected: src_x=4, src_y=0, dest_x=16, dest_y=7.
%Error: /mnt/users/spin0/no_backup/drichmond/Research/repositories/git/bsg_bladerunner.6/bsg_manycore/v/bsg_manycore_link_sif_tieoff.v:45: Verilog $stop
```

dpetrisko:
Okay, this works for 4x2 manycore (VCS/Verilator) with the following patch. This I believe corresponds to the BSG_MACHINE_ORIGIN_COORD_X, BSG_MACHINE_ORIGIN_COORD_Y in machines/pod_X1Y1_ruche_X4Y2_hbm/Makefile.machine.include. But these variables are not exposed by the configuration ROM. Seems like we should add them? I gave it a try but got lost in the ROM magic(tm)

dpetrisko:
Okay, figured out the rom.  Should be good for review

libraries/bsg_manycore_eva.cpp
```diff
@@ -1001,8 +1001,8 @@ int default_eva_size(
 
 }
 
-const hb_mc_coordinate_t default_origin = {.x = HB_MC_CONFIG_VCORE_BASE_X,
-                                           .y = HB_MC_CONFIG_VCORE_BASE_Y};
+const hb_mc_coordinate_t default_origin = {.x = 0,
```
dpetrisko, line 1004: Not sure what the right default is here (what is default_map used for)? Seems dangerous to have hardcoded origins as we see :)

libraries/bsg_manycore_config.cpp
```diff
@@ -135,6 +135,11 @@ int hb_mc_config_init(const hb_mc_config_raw_t raw[HB_MC_CONFIG_MAX],
         idx = raw[HB_MC_CONFIG_DEVICE_HOST_INTF_COORD_Y];
         config->host_interface.y = idx;
 
+        idx = raw[HB_MC_CONFIG_ORIGIN_COORD_X];
+        config->origin.x = idx;
+
+        idx = raw[HB_MC_CONFIG_ORIGIN_COORD_Y];
+        config->origin.y = idx;
```
drichmond, line 142: I would run CHECK_FIELD after each of these. They should probably be >0 and a power of two.

libraries/bsg_manycore_eva.cpp
drichmond: Oof. I would say -1 and -1 but coordinate is an unsigned int. Certainly not 0, 0, because that could lead to a silent failure.

Maybe INT_MAX

libraries/platforms/bigblade-vcs/hardware/dpi_top.sv
```diff
@@ -12,7 +12,9 @@ module replicant_tb_top
    end
     */
    initial begin
+`ifndef VERILATOR
       #0;
+`endif
       
       $display("==================== BSG MACHINE SETTINGS: ====================");
```
drichmond, line 19: I would add the origin parameters to this initial print

platform.mk
```diff
@@ -44,7 +44,7 @@ endif
 # We default to simulating the AWS machine uinsg Synopsys VCS-MX,
 # HOWEVER, if VCS_HOME is not defined then we will assume that
 # VCS/VCS-MX is not installed and try fall-back options
-BSG_PLATFORM ?= bigblade-vcs
+BSG_PLATFORM ?= bigblade-verilator
```
drichmond, line 48: Undo before merge

machines/pod_X1Y1_ruche_X4Y2_hbm/Makefile.machine.include
```diff
@@ -0,0 +1,139 @@
...
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+############################################################################################
+# TIMING_MESH_8_4: this is a 8x4 array with v-caches on top and bottom (a “8x4 pod”).      #
+# Each v-cache is mapped to a separate HBM bank.                                           #
+# There is one HBM channel.                                                                #
+# There is a vanilla, 2D mesh network (no ruche links)                                     #
+# This is probably not a sensible configuration, as the tile to channel ratio is very low. #
+# But for memory bound benchmarks, it may be reasonable to use for iteration.              #
+############################################################################################
```
drichmond, line 35: Change intro comment

libraries/bsg_manycore_eva.cpp
drichmond: The default map is the original EVA map for the manycore. 

This struct may not matter anymore. It may be unrealized legacy code...

drichmond:
This is looking good to me!

Things to finish:

1. Does the Verilog pod coordinate width change with the number of pods? I recall that this is constant in the code, and having it be dynamic in software may not produce the expected behavior with multiple pods
2. Is the private/coordinate field still used in the EVA map? If not, we should probably do some deprecation (though I see no need to remove it)
3. Add origin prints (verilog) and checks (C++)
4. Do we want a C++ profiling target? Do we want a waveform target? Those would be nice to add now, but require varying degrees of effort. 

We should also do a final pass to merge duplicate files. I think compilation.mk in VCS and Verilator are the same, now that we dynamically load a .so file. There may be others, and with some minor effort we might be able to merge exection.mk?

libraries/bsg_manycore_config.cpp
```diff
@@ -191,11 +196,14 @@ int hb_mc_config_init(const hb_mc_config_raw_t raw[HB_MC_CONFIG_MAX],
...
         }
 
-        // Dervived variables from the ROM
-        config->pod_coord_width  = hb_mc_coordinate(3, 4);
+        // Derived variables from the ROM
         config->tile_coord_width = hb_mc_coordinate(
-            config->noc_coord_width.x - config->pod_coord_width.x,
-            config->noc_coord_width.y - config->pod_coord_width.y
+            log2(config->pod_shape.x),
+            log2(config->pod_shape.y)
```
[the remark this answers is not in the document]
dpetrisko: See this discussion:

https://github.com/bespoke-silicon-group/bsg_manycore/pull/547#discussion_r682135288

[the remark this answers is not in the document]
dpetrisko: Basically yes, we are enforcing that tile coord width == pod dimensions

machines/pod_X1Y1_ruche_X4Y2_hbm/Makefile.machine.include
```diff
@@ -0,0 +1,141 @@
+# Copyright (c) 2019, University of Washington All rights reserved.
```
dpetrisko, line 1: Note, this is X8Y4 right now, work needs to be done to support X4Y4 or X4Y2

dpetrisko:
Ready to go. Merge after:
- https://github.com/bespoke-silicon-group/basejump_stl/pull/553
- https://github.com/bespoke-silicon-group/bsg_manycore/pull/635
- https://github.com/bespoke-silicon-group/bsg_manycore/pull/636

drichmond:
This passes Verilator CUDA Regression with the previously mentioned PRs. Library is running slowly.

Will do a final verification that we didn't break VCS before merging.

drichmond:
Passes both regression suites in Verilator and in VCS

mrutt92 (approved):

PR #761, opened 2022-01-25
Add parameters for icache line size

drichmond:


drichmond:
Waiting for SPMD tests to finish, but CUDA and library tests passed

drichmond:
The indefinite tests are not related to this icache change, I believe. I added them in this PR, because they were missing from our regression. They pass in manycore.

mrutt92:
looks good, a couple of thoughts:

First, a large chunk of this PR is the same makefile copied over and over. I think we should either either have a common makefile that is included, or a template makefile that is copied by some script, or some other method to generate these makefiles. The problem with the current approach is, whenever we make some change to our flow (which has been known to happen due to tape outs, experimental needs, other crap), we then need to go and make the same change to an ever increasing list of makefiles.

Second (more of a suggestion), I'm wondering if we should run some of the motifs on this before merging? In the past we experienced issues running very long simulations because of a dangling "inv" parameter that we didn't know needed to be set. We can ignore this if we have a very high confidence that this is not the case here.

mrutt92:
alternatively to the second point, we could caution app developers about this change and tell them to run their shorter jobs first to see if they run into problems with the new release. what i'd like to avoid is people kicking off very long simulations and then seeing some simulator error pop up after having wasted days of simulation time.

drichmond:
Good points

> looks good, a couple of thoughts:
> 
> First, a large chunk of this PR is the same makefile copied over and over. I think we should either have a common makefile that is included, or a template makefile that is copied by some script, or some other method to generate these makefiles. The problem with the current approach is, whenever we make some change to our flow (which has been known to happen due to tapeouts, experimental needs, other crap), we then need to go and make the same change to an ever increasing list of makefiles.

Yeah, I agree. However, the bulk of those makefiles is the copyright statement at the top, a comment, and a few lines of including other makefiles (which may include defining the tile group size in the second case). 

So I agree that there's a lot of copying and there a better way to do it. I think I experimented with automatically generating the makefile during regression and just got frustrated with all the ways that could go wrong. 

Fortunately, it's not a lot of code. Just a lot of boilerplate. Will not fix in this PR, but open to solutions in the future.

> 
> Second (more of a suggestion), I'm wondering if we should run some of the motifs on this before merging? In the past we experienced issues running very long simulations because of a dangling "inv" parameter that we didn't know needed to be set. We can ignore this if we have a very high confidence that this is not the case here.

Yes. Perhaps this should be as an filed as an issue, but this repository needs to switch over to the BSG parameter macros. That would reduce the possibility of an "inv" bug (and was incidentally how I found that we were missing a parameter, and led to this PR).

drichmond:
> alternatively to the second point, we could caution app developers about this change and tell them to run their shorter jobs first to see if they run into problems with the new release. what i'd like to avoid is people kicking off very long simulations and then seeing some simulator error pop up after having wasted days of simulation time.

Fully agree. The CUDA tests run, but that's no guarantee of overall success. @tommydcjung told me that there are no software changes necessary so I'm confident we're OK on that front unless there is a functional bug. The "inv" issue is more concerning. I think that has been somewhat reduced by the macros I mentioned above, though switching to those macros in the replicant RTL would make me more confident.

It's a rote task, but perhaps an easy one for a new student coming up to speed on the project.

For now, I think saying "run some smoketests first to confirm" is fine.

libraries/platforms/bigblade-vcs/hardware/dpi_top.sv
```diff
@@ -26,6 +26,10 @@ module replicant_tb_top
       $display("[INFO][TESTBENCH] bsg_machine_pod_tiles_subarray_y_gp   = %d", bsg_machine_pod_tiles_subarray_y_gp);
       $display("[INFO][TESTBENCH] bsg_machine_pod_llcache_rows_gp       = %d", bsg_machine_pod_llcache_rows_gp);
 
+      $display("[INFO][TESTBENCH] bsg_machine_core_icache_line_words_gp = %d", bsg_machine_core_icache_line_words_gp);
+      $display("[INFO][TESTBENCH] bsg_machine_core_icache_entries_gp    = %d", bsg_machine_core_icache_entries_gp);
```
mrutt92, line 30: where is this parameter used?

drichmond: line_words is used in the top level, it is passed into the testbench.

icache_entries is passed internally via the package it is part of. I added these just to have them printed out.

hardware/hardware.mk
```diff
@@ -288,6 +288,7 @@ $(BSG_MACHINE_PATH)/bsg_bladerunner_pkg.v: $(BSG_MACHINE_PATH)/bsg_bladerunner_c
 	@echo "parameter bsg_machine_core_dmem_words_gp = 1024;" >> $@
 	@echo "parameter bsg_machine_core_icache_entries_gp = 1024;" >> $@
 	@echo "parameter bsg_machine_core_icache_tag_width_gp = 12;" >> $@
+	@echo "parameter bsg_machine_core_icache_line_words_gp = 4;" >> $@
```
mrutt92, line 291: so is `bsg_machine_core_icache_line_words_gp` = `icache_block_size_in_words_p`?

drichmond: Yes?

mrutt92: we had a discussion about naming offline

mrutt92 (approved):

mrutt92:
is this good to merge @drichmond ?

Issue #631, opened 2020-07-31
Request for BSG Bladerunner ASIC Synthesis and Implementation Report


vijayholimath:
@drichmond and the team, may I request you to provide ASIC synthesis and implementation report of BSG Bladerunner? Unfortunately, I don't have Cadence and Synopsys tools.  I am using FPGA version 3.5.2

Many Thanks.

taylor-bsg:
Hi Vijay,

I hope you are well. The numbers (frequency, cores per mm^2) are very
similar to the Celerity chip which is described  off of my webpage.

M

vijayholimath:
Many thanks for the clarification, Michael...:) Yes, I am aware of the paper "The Celerity Open-Source 511-Core RISC-V Tiered Accelerator Fabric" . By the way, we have requirement coming-up to fabricate chip. May I request you to assist us in implementing the chip? Of course, the team/group will be compensated..:)

PR #765, opened 2022-03-04
Matmul blocked

drichmond:
This is our most optimal (but not necessarily optimum) implementation of SGEMM.

There are two versions: 

SGEMM Data Parallel, where each tile computes a 16x16 block

SGEMM Group Cooperative, where each tile works with its NxN neighbors to compute an 16Nx16N block

The latter is more optimized.

Merging this in now to avoid letting it get stale.


drichmond:
This is ready to merge after #769 and [#631](https://github.com/bespoke-silicon-group/bsg_manycore/pull/631)

PR #767, opened 2022-03-04
Black-Scholes

drichmond:
Black Scholes Manycore code. Derived from Parboil benchmark

The most optimized version uses a custom link script to put RODATA in scratchpad for logf/expf functions.

Ready to merge


drichmond:
This is more or less ready to go, but the opt-pod version is producing incorrect results. I think the issue is probably related to indexing, because the single-thread version is correct.

drichmond:
Fixed

PR #769, opened 2022-03-08
Group Strider Test

drichmond:


examples/cuda/group_stride/Makefile
```diff
@@ -0,0 +1,133 @@
+# Copyright (c) 2021, University of Washington All rights reserved.
```
mrutt92, line 1: were we ever going to do that thing where we generate these makefiles?

examples/cuda/group_stride/main.cpp
```diff
@@ -0,0 +1,227 @@
...
+        argp_parse (&argp_path, argc, argv, 0, 0, &args);
+        bin_path = args.path;
+        test_name = args.name;
+
+        //
+        // Define path to binary.
+        // Initialize device, load binary and unfreeze tiles.
+        //
+        hb_mc_device_t device;
+        rc = hb_mc_device_init(&device, test_name, 0);
```
mrutt92, line 57: why not use the CUDA_CALL macro here?

examples/cuda/group_stride/Makefile
drichmond: Maybe

examples/cuda/group_stride/main.cpp
drichmond: Dunno, this was written a long time ago. Might not have been around.

examples/cuda/group_stride/main.cpp
```diff
@@ -0,0 +1,227 @@
...
+                bsg_pr_err("failed to de-initialize device.\n");
+                return rc;
+        }
+
+
+        //************************************************************
+        // Calculate the expected result matrix using host code and
+        // compare the results.
+        // ************************************************************
+        for(int ix = 0; ix < tg_dim.x; ++ix){
```
mrutt92, line 194: these are fine, but we have some nice macros for iterating over coordinates...

examples/cuda/group_stride/main.cpp
mrutt92: how was it written a long time ago? your pushing it now?

drichmond: I just did some rebasing. It used to be part of #765 and then I pulled it out to separate concerns.

drichmond: #765 is actually the genesis of the whole strider idea.

mrutt92: I see. I think good style would be to use the CUDA_CALL macros because it keeps these tests short and easier to read.

drichmond: K. Will fix

mrutt92: having to read too much that does nothing makes me cranky.

drichmond: Fixed

mrutt92 (approved):
pending the other pr in manycore is merged

PR #774, opened 2022-03-24
Remove ORIGIN macros, make ROM depependency

drichmond:
* Added hb_mc_manycore_eva_init function
* Added prints to dpi_top
* Compute NoC Coordinate widths to match RTL

Derived (and slightly modified) from #752

### DIFF
--- a/machine.mk
+++ b/machine.mk
@@ -38,7 +38,7 @@ endif
 
 # To switch machines, simply switch the path of BSG_MACHINE_PATH to
 # another directory with a Makefile.machine.include file.
-BSG_MACHINE_PATH ?= $(BSG_F1_DIR)/machines/pod_X1Y1_ruche_X16Y8_hbm_one_pseudo_channel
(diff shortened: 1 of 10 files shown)


drichmond:
Passes regression

machine.mk
```diff
@@ -38,7 +38,7 @@ endif
 
 # To switch machines, simply switch the path of BSG_MACHINE_PATH to
 # another directory with a Makefile.machine.include file.
-BSG_MACHINE_PATH ?= $(BSG_F1_DIR)/machines/pod_X1Y1_ruche_X16Y8_hbm_one_pseudo_channel
```
dpetrisko, line 41: revert default?

drichmond: Yep, my bad

drichmond: Fixed

dpetrisko (approved):
LGTM

PR #775, opened 2022-03-24
Pod X1Y1, Ruche X8Y4

drichmond:
Add X8Y4 single-pod design, and necessary fixes for DMA and other libraries.

Passes regression. 

Depends on #774, derived from #752


dpetrisko (approved):
Nice!

PR #776, opened 2022-03-24
Deprecate platforms

drichmond:
Opening this for tracking


mrutt92:
looks like this should be merged after another pr?

drichmond:
Yeah, after #752

drichmond:
I need to re-run regression on this. I moved the packet snooper out of AWS and want to make sure we're still good to go.

PR #783, opened 2022-04-16
PC Histogram Executable

mrutt92:
Relies on https://github.com/bespoke-silicon-group/bsg_manycore/pull/644.

We have a PC histogram tool currently that parses a cycle-by-cycle trace file from simulation.
The functionality is very helpful, but it comes at ~3x cost to simulation time and the trace file can grown to GBs in size for even relatively short simulations.

This PR provides a fast-functional PC histogram. The histogram is stored in memory while simulation is running and dumped to a file at the end of simulation. Some level of detail is lost such as the instruction opcode (instead we see either 'instr' or 'fp_instr'). However, the stall type information is retained.

The overhead from using the fast PC histogram was never more than 20% over fast non-profiling simulation and as much as 3x faster than enabling the trace.

Note that this is unsuitable for blood graph generation. You still need to use tracing for that.


drichmond:
Is there any reason this wouldn't work with Verilator?

mrutt92:
I assume not? Can you outline what I need to do to include this in verilator? In bsg_manycore it was obvious.

dpetrisko:
Yes binds work in Verilator

mrutt92:
Ok, then it should work fine.

drichmond:
It's not super complex - I'll make a note to add a comment. 

Bind works in verilator, but it can't be turned off by a parameter. At the moment this means we do two defines: 

`BSG_MACHINE_DISABLE_FOO_BAR` - this is your standard profiler disable macro. It turns into a parameter at the top level in dpi_top.sv.

`VERILATOR_WORKAROUND_DISABLE_FOO_BAR` - This is the verilator workaround, it triggers an ifdef macro in bsg_nonsynth_manycore_testbench

We have both defined at the moment so that we don't start to rely on macros to turn on/off the profilers in VCS. When/if verilator fixes the issue we'll just remove the verilator workaround macros.

Here's the current verilator platform / example: 
https://github.com/bespoke-silicon-group/bsg_replicant/blob/eec2f9d67b81d7675455ac841186ce0244a9172a/libraries/platforms/bigblade-verilator/link.mk#L113

libraries/platforms/bigblade-vcs/link.mk
```diff
@@ -71,14 +71,15 @@ VCS_VSOURCES = $(VHEADERS) $(VSOURCES)
 
 $(BSG_MACHINExPLATFORM_PATH)/debug/simv: VCS_VFLAGS += +plusarg_save +vcs+vcdpluson +vcs+vcdplusmemon +memcbk -debug_pp
 
-$(BSG_MACHINExPLATFORM_PATH)/saifgen/simv $(BSG_MACHINExPLATFORM_PATH)/exec/simv: VDEFINES += BSG_MACHINE_DISABLE_VCORE_PROFILING
-$(BSG_MACHINExPLATFORM_PATH)/saifgen/simv $(BSG_MACHINExPLATFORM_PATH)/exec/simv: VDEFINES += BSG_MACHINE_DISABLE_CACHE_PROFILING
-$(BSG_MACHINExPLATFORM_PATH)/saifgen/simv $(BSG_MACHINExPLATFORM_PATH)/exec/simv: VDEFINES += BSG_MACHINE_DISABLE_ROUTER_PROFILING
```
drichmond, line 76: Did the saifgen rules get removed? Or am I just being confused by git

mrutt92: you're being confused by git. i added pc-histogram as one of the targets for which this rule applies and saifgen is still one of those targets.

drichmond (approved):

libraries/platforms/bigblade-vcs/link.mk
mrutt92: Not sure. Maybe? I figured I'd let this new tool breath a bit before I decided for other people how and when it should be used.

drichmond: That's reasonable to me. 

:ship: it

mrutt92: But I'll happily defer to your opinion...

PR #784, opened 2022-04-20
Call top->finish() during program exit

drichmond:
There were two issues:

1. top->finish() wasn't being called to trigger final blocks in Verilator
2. the deconstructor for SimulationWrapper wasn't being called, so the call site for adding top->finish() wasn't reached

### DIFF
--- a/libraries/platforms/bigblade-verilator/bsg_manycore_platform.cpp
+++ b/libraries/platforms/bigblade-verilator/bsg_manycore_platform.cpp
@@ -162,6 +162,15 @@ void hb_mc_platform_cleanup(hb_mc_manycore_t *mc)
         auto key = active_ids.find(platform->id);
         active_ids.erase(key);
 
+        delete platform->top;
+        auto m = machines.find(platform->id);
+        if(m != machines.end()){
+                machines.erase(m);
+        } else {
+                // Whoa, this would be an error condition but there's really no good way to handle it.
+                // Possible causes: Cleanup before init, memory corruption
+                manycore_pr_err(mc, "Machine ID %d was not found during platform cleanup. Memory corruption?", platform->id);
+        }
         platform->id = 0;
 
         // Ideally, we would clean up each platform in
--- a/libraries/platforms/bigblade-verilator/bsg_manycore_simulator.cpp
+++ b/libraries/platforms/bigblade-verilator/bsg_manycore_simulator.cpp
@@ -69,6 +69,7 @@ std::string SimulationWrapper::getRoot(){
 
 SimulationWrapper::~SimulationWrapper(){
         Vreplicant_tb_top *top = reinterpret_cast<Vreplicant_tb_top *>(this->top);
+        top->final();
         delete top;
         this->top = nullptr;
 }


dpetrisko:
When $finish is called, how does that trigger this destructor? It doesn’t look like Verilated::gotFinish is used as in the example

drichmond:
Hmm. do you have an example of gotFinish?

🐟

dpetrisko:
https://github.com/verilator/verilator/blob/880a9be3b10763b1d0b68414c1fa02700b3377d2/examples/make_hello_c/sim_main.cpp#L28

drichmond:
The current implementation doesn't support calling finish from verilog. We probably should, but it would take some work. 

At the very least we would need to check for it in these locations.

https://github.com/bespoke-silicon-group/bsg_replicant/blob/c21a2b15a9956c91b9a1eaffedcd1fb172ebdc91/libraries/platforms/bigblade-verilator/bsg_manycore_platform.cpp#L611

https://github.com/bespoke-silicon-group/bsg_replicant/blob/c21a2b15a9956c91b9a1eaffedcd1fb172ebdc91/libraries/platforms/bigblade-verilator/bsg_manycore_platform.cpp#L85

https://github.com/bespoke-silicon-group/bsg_replicant/blob/c21a2b15a9956c91b9a1eaffedcd1fb172ebdc91/libraries/platforms/bigblade-verilator/bsg_manycore_platform.cpp#L306

https://github.com/bespoke-silicon-group/bsg_replicant/blob/c21a2b15a9956c91b9a1eaffedcd1fb172ebdc91/libraries/platforms/bigblade-verilator/bsg_manycore_platform.cpp#L350

https://github.com/bespoke-silicon-group/bsg_replicant/blob/c21a2b15a9956c91b9a1eaffedcd1fb172ebdc91/libraries/platforms/bigblade-verilator/bsg_manycore_platform.cpp#L417

https://github.com/bespoke-silicon-group/bsg_replicant/blob/c21a2b15a9956c91b9a1eaffedcd1fb172ebdc91/libraries/platforms/bigblade-verilator/bsg_manycore_platform.cpp#L455

Support may require some sort of return value back up the stack to signal an unrecoverable failure.

dpetrisko:
Oh no that’s fine, I think I understand the flow then. As long as there’s a single point  to determine when the simulation should exit and it funnels to this final call

libraries/platforms/bigblade-verilator/bsg_manycore_platform.cpp
```diff
@@ -162,6 +162,15 @@ void hb_mc_platform_cleanup(hb_mc_manycore_t *mc)
         auto key = active_ids.find(platform->id);
         active_ids.erase(key);
 
+        delete platform->top;
+        auto m = machines.find(platform->id);
```
dpetrisko, line 166: When are there more than one machine?

drichmond: This is legacy, but hasn't been removed. When I wrote it, I was trying to support having having multiple HB instances being simulated.

drichmond:
OK. We _should_ try to support it at some point IMO. I'll file an issue. To me, it's an exceptional/hardware development occurrence, not a software development occurrence.

libraries/platforms/bigblade-verilator/bsg_manycore_platform.cpp
dpetrisko: Okay!

dpetrisko (approved):
LGTM

PR #786, opened 2022-04-21
Fix LDFLAGS

drichmond:
We used to link the pod repl library when we wanted to replicate calls across pods.

Now the library is linked into the simulation executable and is no
longer necessary here. If a program wants to call it, the user should
include the header in the program, and when main.so is dynamically
loaded the appropriate symbols will exist in the execution binary


mrutt92:
I don't get it. How do you choose to swap in the pod replication?

drichmond:
Isn't the API different?

I've never actually used the repl library

mrutt92:
No. The api is the same but we substitute at link time.

drichmond:
Hmm. I need to think about this. I don't think our current simulation build system works with this. 

Do we have a test that would break here?

mrutt92:
No actually: in practice the pod replication stuff might be something we apply to the entire regression instead of individual tests. In which case, these changes would be appropriate and we add a commented out line/option to turn this on in some shared makefile (link.mk?)

drichmond:
We could also create a repl target.

drichmond:
OK, I went ahead and implemented a potential solution. There's now a repl execution target which will use the API in the replicated library. See the most recent commit

drichmond:
@mrutt92 OK with the changes here?

mrutt92 (approved):

PR #782, opened 2022-04-15
Xcelium platform

dpetrisko:
This PR adds an xcelium platform. Should be the last platform for a while ;)

It does the dynamic loading in the same way as VCS and verilator. However there are a few idiosyncrasies:
- dumping has no command line switch in xcelium (please prove me wrong), so we're forced to use a TCL script to enable dumping on dumping configs
- xmsim does not appear to support spaces in plusargs strings. Therefore, we do a workaround where we substitute @ for spaces in the c_args, and then unsubstitute in the C. Open to suggestions for a cleaner approach 

This PR does not add SAIFGEN target.  Goes along with: https://github.com/bespoke-silicon-group/bsg_manycore/pull/643

PR #796, opened 2022-06-21
Consolidating platforms into common/dpi

dpetrisko:
This PR breaks out the common parts of the DPI platforms into a single directory, leaving only the platform-specific stuff.

Depends on #782, the diff will get a lot cleaner when that goes through


libraries/platforms/common/hardware/bsg_print_stat_snoop.v
```diff
@@ -0,0 +1,46 @@
+module bsg_print_stat_snoop
```
drichmond, line 1: I am split about this. The common place for non-platform hardware has always been in the root hardware directory.

dpetrisko: Oh I don’t feel passionately at all. I just moved it so that it would be easier to deprecate AWS-VCS, fine with staging in hardware/ until the move to bsg_manycore is complete (or leaving as is)

libraries/libraries.mk
```diff
@@ -110,7 +110,7 @@ $(BSG_PLATFORM_PATH)/libbsg_manycore_runtime.so.1.0: INCLUDES :=
 
 include $(BSG_PLATFORM_PATH)/library.mk
 
-$(LIB_OBJECTS) $(LIB_OBJECTS_CUDA_POD_REPL) $(LIB_OBJECTS_REGRESSION): INCLUDES := -I$(LIBRARIES_PATH)
+$(LIB_OBJECTS) $(LIB_OBJECTS_CUDA_POD_REPL) $(LIB_OBJECTS_REGRESSION): INCLUDES += -I$(LIBRARIES_PATH)
```
drichmond, line 113: There's a very specific reason that this is :=, not +=. Can you revert this?

libraries/platforms/common/hardware/bsg_print_stat_snoop.v
drichmond: Let's move to hardware/

libraries/libraries.mk
dpetrisko: Yep, thought I caught all of those

libraries/platforms/bigblade-xcelium/execution.mk
```diff
@@ -0,0 +1,95 @@
...
+exec.log: $(BSG_MACHINE_PATH)/$(BSG_PLATFORM)/exec/simx
+profile.log: $(BSG_MACHINE_PATH)/$(BSG_PLATFORM)/profile/simx
+pc-histogram.log: $(BSG_MACHINE_PATH)/$(BSG_PLATFORM)/pc-histogram/simx
+repl.log: $(BSG_MACHINE_PATH)/$(BSG_PLATFORM)/repl/simx
+
+# Use VANILLA_STATS and VCACHE_STATS to override paths
+VANILLA_STATS ?= vanilla_stats.csv
+VCACHE_STATS ?= vcache_stats.csv
+stats: profile.log
+	PYTHONPATH=$(BSG_MANYCORE_DIR)/software/py/ python3 -m vanilla_parser --only stats_parser --stats $(VANILLA_STATS) --vcache-stats $(VCACHE_STATS)  --tile-group --tile --cache-line-words $(BSG_MACHINE_VCACHE_LINE_WORDS)
```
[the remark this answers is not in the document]
dpetrisko: spilt out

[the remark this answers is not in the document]
drichmond: Did you catch the one in riscv.mk too (I didn't tell you about it but someone else ran into it this week)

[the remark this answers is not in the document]
dpetrisko: Removed!

dpetrisko:
This is super duper old, but I did rebase and it's clean so we can get it merged whenever's convenient

drichmond (approved):

dpetrisko:
Thanks!

PR #797, opened 2022-07-15
Pagerank 6.7.1

zzy82518996:
This PR add the pagerank example.
The code base contains:
1. graphit: constains the host code that handling different preprocessing methods over graphs.
2. kernels: contains different optimization strategies of kernels in pagerank.
3. Other folders: contains the host code and scipt of running different optimization strategies.


zzy82518996:
> @zzy82518996 do we need the graphit code to compile and run your implementation?

Yes, there are some modifications on the code inside graphit. But most of the modified code corresponds to HB host.

mrutt92:
what are the modifications to graphit? can you send me a diff? i'm a bit hesitant to just merge a copy of the graphit source code into this repo. the way we usually handle this at UW is we include a rule for cloning the subrepo in a makefile.

which graphit for is this source code from? i imagine it would be the one associated with BSG?

zzy82518996:
I will send you a diff and yes you are right, the graphit code I use is the one associated with BSG.

PR #808, opened 2022-11-10
Adding atomic packet example

dpetrisko:
This PR adds the ability to do AMO.ADDs from host to manycore DRAM space. This can be useful for synchronization between host and device

Question: Is there a low-overhead function to detect that an NPA is a DRAM address? Would like to add the check to amoadd because I believe sending an AMOADD to a vcore tile could cause a hang.

### DIFF
--- a/libraries/bsg_manycore.cpp
+++ b/libraries/bsg_manycore.cpp
@@ -1310,6 +1310,10 @@ int hb_mc_manycore_amoadd(hb_mc_manycore_t *mc, const hb_mc_npa_t *npa, const ui
         if (err != HB_MC_SUCCESS)
                 return err;
 
+        // AMO are only supported on DRAM regions
+        if (hb_mc_config_is_dram(hb_mc_manycore_get_config(mc), hb_mc_npa_get_xy(npa)) == 0)
+            return HB_MC_INVALID;
+
         hb_mc_request_packet_set_op(&rqst.request, HB_MC_PACKET_OP_REMOTE_AMOADD);
         hb_mc_request_packet_set_data(&rqst.request, v);
 
--- a/libraries/bsg_manycore.cpp
+++ b/libraries/bsg_manycore.cpp
@@ -1284,6 +1284,58 @@ int hb_mc_manycore_write32(hb_mc_manycore_t *mc, const hb_mc_npa_t *npa, uint32_
         return hb_mc_manycore_write(mc, npa, &v, 4);
 }
 
+/**
+ * Do a 32-bit amoadd to manycore hardware at a given NPA (must be a DRAM address)
+ * @param[in]  mc     A manycore instance initialized with hb_mc_manycore_init()
+ * @param[in]  npa    A valid hb_mc_npa_t aligned to a four byte boundary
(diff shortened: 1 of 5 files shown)


mrutt92:
Check in some of the config headers. Grepping for "is_dram" should turn something up.

libraries/bsg_manycore.cpp
```diff
@@ -1284,6 +1284,59 @@ int hb_mc_manycore_write32(hb_mc_manycore_t *mc, const hb_mc_npa_t *npa, uint32_
...
+
+        /* format the request packet */
+        err = hb_mc_manycore_format_request_packet(mc, &rqst.request, npa);
+        if (err != HB_MC_SUCCESS)
+                return err;
+
+        hb_mc_epa_t epa = hb_mc_npa_get_epa(npa);
+
+        // AMO are always 4B
+        size_t sz = 4;
```
mrutt92, line 1308: it's been a while since i've lived in this code base. do we hard code '4' much? maybe we should make a macro 'FOUR' :-D

libraries/bsg_manycore.h
```diff
@@ -253,6 +253,17 @@ extern "C" {
...
+        /**
+         * Do a 32-bit amoadd to manycore hardware at a given NPA (must be a DRAM address)
+         * @param[in]  mc     A manycore instance initialized with hb_mc_manycore_init()
+         * @param[in]  npa    A valid hb_mc_npa_t aligned to a four byte boundary
+         * @param[in]  v      A word value to be added to the NPA
+         * @param[out] vpo    Pointer to the previous value at the NPA
+         * @return HB_MC_SUCCESS on success. Otherwise an error code defined in bsg_manycore_errno.h.
+         */
+        __attribute__((warn_unused_result))
+        int hb_mc_manycore_amoadd(hb_mc_manycore_t *mc, const hb_mc_npa_t *npa, uint32_t vpi, uint32_t *vpo);
```
mrutt92, line 265: in keeping with naming conventions, i request this be named amoadd32

libraries/bsg_manycore.cpp
dpetrisko: Indeed we do :)
```
int hb_mc_manycore_write32(hb_mc_manycore_t *mc, const hb_mc_npa_t *npa, uint32_t v)
{
        return hb_mc_manycore_write(mc, npa, &v, 4);
}
```

libraries/bsg_manycore.h
dpetrisko: done

dpetrisko:
>Check in some of the config headers. Grepping for "is_dram" should turn something up.

Found it, thanks!

libraries/bsg_manycore.cpp
mrutt92: haha, well that's very slightly unfortunate, but hey whatever.

mrutt92: since it's now embedded in the name of the function that this is a 32-bit op, it seems more OK.

mrutt92 (approved):
Lgtm

PR #826, opened 2023-07-29
ZynqParrot FPGA Platform

dpetrisko:
This PR adds the ZynqParrot FPGA platform for use with ZynqParrot + Manycore. This is backwards compatible with the current ps.cpp-based flow but allows for running CUDA-lite programs instead of bare-metal manycore programs


dpetrisko:
Depends on #825

PR #829, opened 2023-12-21
Updates for latest manycore

dpetrisko:
This PR brings bsg_replicant up to speed with BaseJump 2.0 and ipoly changes. Note that all machine configs have disabled ipoly because the CUDA library must be modified to support it. In the meantime, legacy machines continue to be supported


mrutt92 (approved):

dpetrisko:
Thanks!

PR #834, opened 2024-11-12
Enable ipoly hashing, add flags, enable periodic stats, support verdi, add submodule, minor fixes

### LINKED ISSUE #820: Running on Rocky 8.7
Hi I'm looking at this commit 6515c42

1.
I notice that in [examples/spmd/Makefile](https://github.com/bespoke-silicon-group/bsg_replicant/blob/main/examples/spmd/Makefile#L73), we kind of assume that all the examples appearing in bsg_manycore would also exist in bsg_replicant. Would it be better if we can just search under bsg_replicant/examples/spmd/ so that if there are missing examples, we won't have "No such file or directory" error?

2.
In [libraries/bsg_manycore_features.h](https://github.com/bespoke-silicon-group/bsg_replicant/blob/main/libraries/bsg_manycore_features.h#L33), we test if _BSD_SOURCE is defined. However, in newer glibc versions, _BSD_SOURCE is deprecated and is replaced by _DEFAULT_SOURCE according to https://stackoverflow.com/questions/29201515/what-does-d-default-source-do. Without the following changes I wouldn't be able to run make regression on Rocky 8.7, so perhaps we can update this as well.
```
#if !defined _BSD_SOURCE && !defined _DEFAULT_SOURCE
#error "BSD functionality required for bsg_manycore_runtime"
```

If these changes make sense I can make a PR. Thanks!

### LINKED ISSUE #821: VCS Parsing Error
The commit I am looking at: [6515c42](https://github.com/bespoke-silicon-group/bsg_replicant/commit/6515c42b03730a1f2e584d81612976a4615d0ac8)

With VCS 2022, when compiling pod_X1Y1_ruche_X16Y8_hbm_one_pseudo_channel/bigblade-vcs/exec/simv during "make regression", I will get the following error:

> Source file "tb_timescale=1ps/1ps" cannot be opened for reading due to 'No such file or directory'.

Looks like VCS thinks "tb_timescale=1ps/1ps" is a filename instead of some parameter. Not sure if it is a bug from 2022 VCS.

https://github.com/bespoke-silicon-group/bsg_replicant/blob/65a2398c3f256370f4f315bd12f6fa9c89deee3a/libraries/platforms/bigblade-vcs/link.mk#L61

### LINKED ISSUE #830: VCS option "-ntb_opts tb_timescale=1ps/1ps" doesn't appear to work in vcs/U-2023.03-SP2-8
As part of the vcs compile options, when "-ntb_opts tb_timescale=1ps/1ps" is specified, I get the following error

```
Error-[SFCOR] Source file cannot be opened
  Source file "tb_timescale=1ps/1ps" cannot be opened for reading due to 'No 
  such file or directory'.
  Please fix above issue and compile again.
```

https://github.com/bespoke-silicon-group/bsg_replicant/blob/d91d1e9df979a62b46d14edd2e03393360810351/libraries/platforms/bigblade-vcs/link.mk#L63C1-L64C1

Removing that option allows the simulation to proceed but I am unsure what the impact is

### LINKED ISSUE #833: Missing pc_histogram related files in Makefiles
[hardware.mk](https://github.com/bespoke-silicon-group/bsg_replicant/blob/main/libraries/platforms/common/dpi/hardware.mk#L92) line 92 missing `VSOURCES += $(BSG_MANYCORE_DIR)/testbenches/common/v/vanilla_core_pc_histogram.sv`
[library.mk](https://github.com/bespoke-silicon-group/bsg_replicant/blob/main/libraries/platforms/common/dpi/library.mk#L35) line 35 missing `include $(LIBRARIES_PATH)/features/pc_histogram/simulation/feature.mk`

Currently `make debug.log` is broken when running with hb_hammerbench

These two lines were accidentally dropped in PR #796 when consolidating platforms into common/dpi.

gaozihou:
This is an essential step to keep the main branch up to date

Fixes #820 
Fixes #821 
Fixes #830
Fixes #833 

Note that according to both VCS and VCS-MX user guides, -M is not a valid flag, -L filename is the logging file specification flag. None of these flags should show up in our link Makefile.


tommydcjung (approved):

libraries/features/dma/simulation/dramsim3.mk
```diff
@@ -64,6 +64,9 @@ $(LIBRARIES_PATH)/features/dma/simulation/libdramsim3.so: CXXFLAGS += -I$(BASEJU
 $(LIBRARIES_PATH)/features/dma/simulation/libdramsim3.so: CXXFLAGS += -I$(BASEJUMP_STL_DIR)/imports/DRAMSim3/ext/fmt/include
 $(LIBRARIES_PATH)/features/dma/simulation/libdramsim3.so: CXXFLAGS += -DFMT_HEADER_ONLY=1
 $(LIBRARIES_PATH)/features/dma/simulation/libdramsim3.so: CXXFLAGS += -DBASEJUMP_STL_DIR="$(BASEJUMP_STL_DIR)"
+$(LIBRARIES_PATH)/features/dma/simulation/libdramsim3.so: CXXFLAGS += -DBLOOD_GRAPH
```
dpetrisko, line 67: remove comments if deprecated?

libraries/bsg_manycore_features.h
```diff
@@ -29,9 +29,9 @@
 #define BSG_MANYCORE_FEATURES_H
 // <features.h> sorts out many of these defines based on compile time flags (e.g. -std=c++11)
 #include <features.h>
-// check _BSG_SOURCE
-#ifndef _BSD_SOURCE
-#error "_BSD_SOURCE not defined: required for bsg_manycore_runtime"
+// check _DEFAULT_SOURCE
```
dpetrisko, line 32: IIRC _DEFAULT_SOURCE and _BSD_SOURCE should both be defined

libraries/platforms/common/dpi/hardware/dpi_top.sv
```diff
@@ -12,6 +12,13 @@ module replicant_tb_top
    end
 `endif
 
+`ifdef ENABLE_FSDB_DUMP
```
dpetrisko, line 15: is there no +vcs+fdsbdumpon or similar?

dpetrisko (approved):
LGTM! Minor questions

libraries/features/dma/simulation/dramsim3.mk
gaozihou: @tommydcjung I guess these comments might be intentional..? We are expecting users to uncomment these lines if they are trying to generate blood graphs.. By default we disable them to reduce the amount of data dumped right?

libraries/platforms/common/dpi/hardware/dpi_top.sv
gaozihou: I tried to search for such a flag on Google but failed.. I guess Synopsys's instruction was dumping in verilog code: https://www.youtube.com/watch?v=5qvB4nUCV8U

libraries/features/dma/simulation/dramsim3.mk
tommydcjung: Yes, for now, this is how we turn off blood graph... maybe add a comment explaining that?

libraries/platforms/common/dpi/hardware/dpi_top.sv
dpetrisko: kk!

libraries/bsg_manycore_features.h
gaozihou: Good catch! Yeah we need both to support both `glibc 2.17 (centos 7)` and `glibc 2.28 (rocky 8)`. The `glibc` version can be checked with `ldd --version` command.

I think a more appropriate fix is to add -D_BSD_SOURCE to `libraries/platforms/common/dpi/library.mk`, which is the root cause of this error, @dpetrisko what do you think?

libraries/platforms/common/dpi/library.mk
```diff
@@ -50,8 +50,8 @@ $(PLATFORM_OBJECTS) $(PLATFORM_REGRESSION_OBJECTS): INCLUDES += -I$(BSG_PLATFORM
 $(PLATFORM_OBJECTS) $(PLATFORM_REGRESSION_OBJECTS): INCLUDES += -I$(BSG_MANYCORE_DIR)/testbenches/dpi/
 $(PLATFORM_OBJECTS) $(PLATFORM_REGRESSION_OBJECTS): INCLUDES += -I$(BASEJUMP_STL_DIR)/bsg_test/
 
-$(PLATFORM_OBJECTS) $(PLATFORM_REGRESSION_OBJECTS): CFLAGS    = -std=c11 -fPIC -D_GNU_SOURCE -D_DEFAULT_SOURCE $(INCLUDES)
-$(PLATFORM_OBJECTS) $(PLATFORM_REGRESSION_OBJECTS): CXXFLAGS  = -std=c++11 -fPIC -D_GNU_SOURCE -D_DEFAULT_SOURCE $(INCLUDES)
+$(PLATFORM_OBJECTS) $(PLATFORM_REGRESSION_OBJECTS): CFLAGS    = -std=c11 -fPIC -D_GNU_SOURCE -D_BSD_SOURCE -D_DEFAULT_SOURCE $(INCLUDES)
+$(PLATFORM_OBJECTS) $(PLATFORM_REGRESSION_OBJECTS): CXXFLAGS  = -std=c++11 -fPIC -D_GNU_SOURCE -D_BSD_SOURCE -D_DEFAULT_SOURCE $(INCLUDES)
```
gaozihou, line 54: @dpetrisko I guess this fix is better? Adding both flags to the flow and checking both in our manycore header file?

libraries/features/dma/simulation/dramsim3.mk
gaozihou: Comments added!

libraries/platforms/common/dpi/library.mk
dpetrisko: works for me!

hardware/hardware.mk
```diff
@@ -249,6 +251,8 @@ $(BSG_MACHINE_PATH)/bsg_bladerunner_pkg.sv: $(BSG_MACHINE_PATH)/bsg_bladerunner_
 	@echo >> $@
 	@echo "parameter int bsg_machine_pod_tiles_y_gp = $(BSG_MACHINE_POD_TILES_Y);" >> $@
 	@echo "parameter int bsg_machine_pod_tiles_x_gp = $(BSG_MACHINE_POD_TILES_X);" >> $@
+	@echo "\`define BSG_MACHINE_GLOBAL_Y $(BSG_MACHINE_POD_TILES_Y)" >> $@
+	@echo "\`define BSG_MACHINE_GLOBAL_X $(BSG_MACHINE_POD_TILES_X)" >> $@
```
gaozihou, line 255: @tommydcjung Would it be okay to remove these two defines? Seems like they are repetitive, we already defined them in vcs run here: https://github.com/bespoke-silicon-group/bsg_replicant/blob/main/libraries/platforms/bigblade-vcs/link.mk#L52-L53

I tried removing these two lines and the memcpy test is running fine, `BSG_MACHINE_GLOBAL_Y` and `BSG_MACHINE_GLOBAL_X` are still defined in vcs simulation. Is there anything that I didn't notice?

machines/pod_X1Y1_ruche_X16Y8_hbm_one_pseudo_channel/Makefile.machine.include
```diff
@@ -71,7 +71,7 @@ BSG_MACHINE_VCACHE_WAY                = 8
 BSG_MACHINE_VCACHE_LINE_WORDS         = 16
 BSG_MACHINE_VCACHE_STRIPE_WORDS       = $(BSG_MACHINE_VCACHE_LINE_WORDS)
 BSG_MACHINE_VCACHE_WORD_TRACKING      = 1
-BSG_MACHINE_VCACHE_IPOLY_HASHING      = 0
+BSG_MACHINE_VCACHE_IPOLY_HASHING      = 1
```
gaozihou, line 74: @tommydcjung Would you suggest enabling ipoly hashing by default in all non-bigblade machine files?

dpetrisko: I suggest keeping the 2x2 mesh as a machine with everything minimal (no ipoly, etc). We don’t want a cross-product, but it can be a machine that has most things “turned off”

tommydcjung: Currently, I-poly is implemented for a pod that is 16 or 32 tiles wide. I would enable it for pod_X1Y1_ruche_X16Y8_hbm_one_pseudo_channel, which is the baseline configuration.

hardware/hardware.mk
tommydcjung: I think it's used for some bind modules that needs the parameter BSG_MACHINE_GLOBAL_Y/X, but the parent module does not have access to. If you think it's repetitive, you can remove it.

PR #836, opened 2024-11-16
Fix c++ initializer in GCC version >=11

gaozihou:
Currently when compiling the libraries with GCC version >=11, it shows the following error:

```
bsg_bladerunner/bsg_replicant/libraries/bsg_manycore_request_packet_id.h:125:40: error: no matching function for call to ?request_packet_id::request_packet_id(<brace-enclosed initializer list>, <brace-enclosed initializer list>, <brace-enclosed initializer list>, int)?
  125 |         request_packet_id(addr, x, y, 1)
```

I tested the fix (c++11 uniform initialization) on CentOS 7, Rocky 8 and 9, with GCC 4, 7, 8, 9 ,10, 11, 12 and 13, fine with c++11, 14, 17 and 20

This fix would be helpful when porting our flow onto Rocky 9 in the future

### DIFF
--- a/libraries/bsg_manycore_request_packet_id.h
+++ b/libraries/bsg_manycore_request_packet_id.h
@@ -95,18 +95,33 @@ extern "C" {
         } hb_mc_request_packet_id_t;
 
         /* this is used to match an address under a mask */
+#ifdef __cplusplus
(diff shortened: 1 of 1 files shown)


dpetrisko (approved):
LGTM!

tommydcjung (approved):

PR #837, opened 2024-11-16
Add missing -D_BSD_SOURCE and -D_DEFAULT_SOURCE flags

gaozihou:
-D_BSD_SOURCE flag is needed by glibc version < 2.19 (CentOS 7)
-D_DEFAULT_SOURCE flag is needed by glibc version >= 2.19 (Rocky 8, 9)

We need both flags to support all platforms

Reference: glibc 2.19 release note https://lists.gnu.org/archive/html/info-gnu/2014-02/msg00001.html


dpetrisko (approved):
LGTM~

PR #838, opened 2024-11-22
Device ID Support

gaozihou:
It is essential to support non-hardcoded device ID to choose among PCIe devices

There should be zero side effect on the current simulation flows, since the default behavior is still passing `DEVICE_ID=0x0` into the flow, so all behaviors should remain unchanged.


dpetrisko (approved):
LGTM!

PR #839, opened 2024-11-22
Enable DRAM Support

gaozihou:
It is essential to add this flag, so that some programs (like spmd tests) can still run when only a subset of launched pods have DRAM attached.

---------

Co-authored-by: Tommy Jung <tommy.dc.jung@gmail.com>


machines/bigblade_pod_X1Y1_ruche_X16Y8_hbm/Makefile.machine.include
```diff
@@ -110,6 +110,10 @@ BSG_MACHINE_IO_EP_CREDITS             = 32
 BSG_MACHINE_IO_HOST_CREDITS           = $(BSG_MACHINE_IO_EP_CREDITS)
 BSG_MACHINE_IO_REMOTE_LOAD_CAP        = $(BSG_MACHINE_IO_EP_CREDITS)
 
+# Enable DRAM pod dimensions
+BSG_MACHINE_ENABLE_DRAM_PODS_X        = $(BSG_MACHINE_PODS_X)
+BSG_MACHINE_ENABLE_DRAM_PODS_Y        = $(BSG_MACHINE_PODS_Y)
```
tommydcjung, line 115: I think there should be a comment describing how these parameters work currently.

gaozihou: Thanks for the suggestions @tommydcjung !

- I've added comments to all the machine files for the usage of this new flag
- I've updated the README.md under machines directory (similar to the manycore README.md) with more detailed information about these flags

gaozihou: @tommydcjung Do you think we should change the flag name to `BSG_MACHINE_ENABLE_DRAM_PROGRAM_LOADING_PODS_X/Y`?

tommydcjung (approved):

libraries/bsg_manycore_config_pod.h
```diff
@@ -62,7 +62,9 @@ extern "C" {
         hb_mc_config_vcore_to_pod(const hb_mc_config_t *cfg, hb_mc_coordinate_t vcore)
         {
                 hb_mc_coordinate_t tile_w = hb_mc_config_tile_coord_width(cfg);
-                return hb_mc_coordinate(vcore.x >> tile_w.x, vcore.y >> tile_w.y);
+                hb_mc_coordinate_t og = hb_mc_coordinate(1 << tile_w.x, 1 << tile_w.y);
+                return hb_mc_coordinate( ((vcore.x - og.x) >> tile_w.x),
+                                         ((vcore.y - og.y) >> tile_w.y)/2 );
```
tommydcjung, line 67: was this a bug before? never seen this one before...

gaozihou: Thanks @tommydcjung ! Yes this was a bug, but I ran `grep` in both `bsg_replicant` and `hb_hammerbench` repos and seems like no one was using this function call before..

PR #842, opened 2024-11-24
SPMD Loader Add Flags for FPGA Tests

gaozihou:
Add useful flags to support FPGA tests, should have no side effect on the existing flows
- Select how many pods to launch at runtime
- Specify if all tiles will send finish packets back
- [BUG FIX] Return error code when only some pods are failing (previously tests would pass silently)

Testings:
- VCS simulation spmd hello test running fine
- New loader.c and loader.mk running fine on the FPGA

---------

Co-authored-by: Tommy Jung <tommy.dc.jung@gmail.com>


Yuan-Mao:
Hey the code looks great! I was able to run it on kk8 but I would need to slightly change some Makefile to let the read bsg_cadenv properly.

Also, some minor C issues I wanted to point out:

* In spmd/loader.c at line [119](https://github.com/bespoke-silicon-group/bsg_replicant/blob/spmd_fpga_support/examples/spmd/loader.c#L129), there are two comparisons between signed and unsigned integers. The current code is fine, but this practice is error-prone as the signed integers will always be converted by compiler to unsigned before the comparisons happen.
* Another thing is, at line 245 in the same file, since we are printing unsigned integers, the format specifier should be %u instead of %d.

gaozihou:
Thanks @Yuan-Mao ! 

- Both comments have been addressed
- I've created issue #856 and will try to address the cadenv problem in another PR

Yuan-Mao (approved):

PR #849, opened 2024-11-26
hw barrier machine variable

tommydcjung:


gaozihou (approved):
LGTM!

- hw barrier default off if BSG_MACHINE_ENABLE_HW_BARRIER is undefined
- bigblade and 2x2-mesh machines hw barrier=off, other machines hw barrier=on
- Both hw barrier on/off work as expected

PR #860, opened 2025-02-04
Fix llvm bin path

Yuan-Mao:
The current llvm path seems to be wrong.


gaozihou (approved):
This bug occurs after merging [bsg_manycore PR 725](https://github.com/bespoke-silicon-group/bsg_manycore/pull/725).

With this PR merged, users will have to rebuild their bsg_manycore riscv toolchain to continue using CLANG

PR #866, opened 2025-06-14
fixes bug for multi-pod tile group eva reading

mrutt92:
* Fixes a bug when translating tile-group EVAs from the host
* Before this commit, no matter what the source tile is the EVA is translated as if the source is on pod (0,0)
* After, the pod of the source core is considered when calculating the coordinate.

### DIFF
--- a/libraries/bsg_manycore_eva.cpp
+++ b/libraries/bsg_manycore_eva.cpp
@@ -77,6 +77,32 @@
 #define DEFAULT_DRAM_BITIDX 31
 #define DEFAULT_DRAM_BITMASK (1ULL << DEFAULT_DRAM_BITIDX)
 
+static uint32_t default_vcore_max_x_coord(const hb_mc_config_t *cfg, const hb_mc_coordinate_t *tgt)
+{
+    hb_mc_dimension_t pods = cfg->pods;
+    hb_mc_coordinate_t pod = {pods.x-1, pods.y-1};
+    hb_mc_coordinate_t og = hb_mc_config_pod_vcore_origin(cfg, pod);
+    return og.x + cfg->pod_shape.x-1;
+}
+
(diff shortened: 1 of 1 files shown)


libraries/bsg_manycore_eva.cpp
```diff
@@ -270,7 +297,7 @@ static int default_eva_to_npa_group(const hb_mc_config_t *cfg,
                 return rc;
         *npa = hb_mc_epa_to_npa(hb_mc_coordinate(x,y), epa);
 
-        bsg_pr_dbg("%s: Translating EVA 0x%08" PRIx32 " for tile (x: %d y: %d) to NPA {x: %d y: %d, EPA: 0x%08" PRIx32 "}. \n",
+        bsg_pr_info("%s: Translating EVA 0x%08" PRIx32 " for tile (x: %d y: %d) to NPA {x: %d y: %d, EPA: 0x%08" PRIx32 "}. \n",
```
gaozihou, line 300: Maybe change back to `bsg_pr_dbg`

libraries/bsg_manycore_eva.cpp
```diff
@@ -77,6 +77,32 @@
 #define DEFAULT_DRAM_BITIDX 31
 #define DEFAULT_DRAM_BITMASK (1ULL << DEFAULT_DRAM_BITIDX)
 
+static uint32_t default_vcore_max_x_coord(const hb_mc_config_t *cfg, const hb_mc_coordinate_t *tgt)
```
gaozihou, line 80: I guess the min-max x-y coordinates should be within current pod, like this? https://github.com/bespoke-silicon-group/bsg_replicant/blob/main/libraries/bsg_manycore_eva.cpp#L343-L356

gaozihou (approved):

